PR #6, opened 2025-08-08
Attestation policy & TPM values

Jakob-Naucke:
Create attestation policy and set TPM reference values.

Includes some plumbing to open port 8080 on Kind host and to use custom images until everything is merged in Trustee.

@alicefr the reference value provision isn't very scalable, but I was under the impression that this would need to be changed anyhow. We could also have a file in the container image or a config map that we mandate already exists.

### DIFF
--- a/operator/src/resource.rego
+++ b/operator/src/resource.rego
@@ -1,0 +1,9 @@
+package policy
+import rego.v1
+
+default allow := false
+
+allow if {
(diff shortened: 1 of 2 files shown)


Jakob-Naucke:
Can un-draft after https://github.com/confidential-containers/trustee-operator/pull/99

PR #7, opened 2025-08-14
Node register service

alicefr:
When a node comes up, should contact the register endpoint to retrieve its identifier. The identifier will be use in the follow setups during the boot phase. For example, as identifier for the LUKs key or workload identifier for the SPIRE agent


alicefr:
The main advantages of this approach are:
1. The node registration ID is independent from any cluster objects and controlled directly by the operator:
    * It is stable, it doesn't change over the node lifetime
    * It is created before any attestation requests or resource retrivial, we are sure we can rely on it to associate the node and the resource coupled with the node and managed by the operator
    * It is unique, we generate a UUID per node registration
 1. The node contacts an endpoint owned by the operator, hence it can extrapolate the IP of the http request and associate it with the ID and be used to identify the Machine object which has this IP
 1. The registration API is defined by the operator and independent from Trustee

As far as it regard the first advantage, the initial idea was to use the IP (or the FQDN) of the node. However, this has
several critical issues.
1. The IP might change over the node lifetime
2. A node could have multiple IPs, hence the HTTP request might come from the same node but from a second IP
3. In the cluster the IP is available on the [Machine](https://cluster-api.sigs.k8s.io/developer/core/controllers/machine)
object from [cluster api](https://cluster-api.sigs.k8s.io/). As it is documented in the [machine controller](https://cluster-api.sigs.k8s.io/developer/core/controllers/machine), the addresses aren't mandatory and they come at a unspecified point in time when the machine is ready. In the case of GCP and Azure the IPs are set at machine creation, but for the KubeVirt CAPI provider the IP is set only when the provider is able to ssh in the node. In the case of kubevirt, waiting for the IP to appear in the machine object will cause a deadlock

An other option that was propose was to rely on the provider ID and rely on [metadata afterburn](https://coreos.github.io/afterburn/usage/attributes/). This value is mandatory for CAPI, however, it has the same timing issue as the addresses. Additionally, the provider ID can be constructed from inside the node only for certain providers. For example, Azure provides too few metadata to be able to construct the provider ID from within the node.

About the second point, for now the attestation protocol interests only the Trustee endpoint, hence it is impossible for the operator to fetch the IP from an attestation request. It would need to infer the IP from somewhere else, like the Machine object, but the information might not be available. Having available the IP from the registration request, the operator is able to match the ID with a machine object, even if this information is available after the firstboot. The IP of the request can be used also for running DNS requests and getting the information that where proposed to be used as identifier. Eventually, we could also include in the registration request additional data like the afterburn metadata.

The registration endpoint has also the advantage that could be use to provide additional metadata to the node then the ID. For example, we could choose to provide the Trustee URL using the metadata instead of setting it using ignition and the clevis pin. Please, note that the clevis pin will still be necessary in order to pass the registration endpoint. 
An additional information that could be pass with the registration endpoint is the url for the SPIRE server when we want to integrate with SPIFFE/SPIRE.

The registration node ID can also be measured and included in the attestation quote.

Please, note that the registration node ID isn't meant to be an authentication mechanism, for this we should rely on a SPIFFE ID; but is rather a mechnism to identify the associated element to the node. In the case of the SPIFFE ID, in order to generate it the SPIFFE sever requires a workload ID and in our case this role can be played by the node registration ID.

/cc @travier @uril

PR #10, opened 2025-08-18
operator: Prepare to generate secret on demand

Jakob-Naucke:
Create function to create a secret for LUKS passphrase per machine on demand. Still called for one generic machine until we have machine identification.

### DIFF
--- a/operator/src/trustee.rs
+++ b/operator/src/trustee.rs
@@ -1,8 +1,9 @@
 use anyhow::anyhow;
 use base64::{Engine as _, engine::general_purpose};
 use crds::{KbsConfig, KbsConfigSpec, Trustee};
+use json_patch::{AddOperation, PatchOperation, TestOperation};
 use k8s_openapi::api::core::v1::{ConfigMap, Secret};
-use kube::api::PostParams;
+use kube::api::{Patch, PatchParams, PostParams};
 use kube::{Api, Client, Error};
 use log::info;
 use openssl::pkey::PKey;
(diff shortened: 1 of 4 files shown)


operator/src/trustee.rs
```diff
@@ -181,28 +182,50 @@ pub async fn generate_reference_values(
...
-        k8s_openapi::ByteString(b"test".to_vec()),
-    );
+pub async fn generate_secret(
+    client: Client,
+    namespace: &str,
+    kbs_config_name: &str,
+    id: &str,
+) -> anyhow::Result<()> {
+    let mut pass = [0; 32];
+    openssl::rand::rand_bytes(&mut pass)?;
```
alicefr, line 192: nit: abstract this into a function

operator/src/trustee.rs
```diff
@@ -181,31 +182,73 @@ pub async fn generate_reference_values(
...
 
+    let kbs_configs: Api<KbsConfig> = Api::namespaced(client, namespace);
+
+    let existing_secrets = kbs_configs
+        .get(kbs_config_name)
+        .await?
+        .spec
+        .kbs_secret_resources;
+    if existing_secrets.iter().any(|s| s == id) {
+        return Err(anyhow!("Secret with ID {id} already present"));
```
alicefr, line 226: Not sure, we really need to error here. It depends on how you react to this error, but I would simply return without error

operator/src/trustee.rs
```diff
@@ -181,31 +182,73 @@ pub async fn generate_reference_values(
...
 }
 
-// TODO: this function needs to be removed, right now it is only for testing the resource
-pub async fn generate_secret(client: Client, namespace: &str, name: &str) -> anyhow::Result<()> {
-    let mut data = BTreeMap::new();
-    data.insert(
-        "key1".to_string(),
-        k8s_openapi::ByteString(b"test".to_vec()),
-    );
+fn random_bytes() -> anyhow::Result<[u8; 32]> {
```
alicefr, line 185: nit: I meant just to call this function `generate_luks_key`

operator/src/trustee.rs
Jakob-Naucke: updated both

alicefr:
Looks good!
/lgtm

PR #12, opened 2025-08-21
Compute PCRs

Jakob-Naucke:
Run job to run confidential-clusters/compute-pcrs. Supersedes the reference value input workflow. Hard-coded for FCOS 42. Supersedes #11.

@alicefr
- Permission might be overly permissive -- any hints or is this fine?
- We could move to an official registry for the compute-pcrs image after merge of https://github.com/confidential-clusters/compute-pcrs/pull/19.

cc @bgartzi


alicefr:
>> add a new binary which launches the calculation and then update the config maps of trustee? I think it will be a cleaner

>Hmm, I have two concerns about it being cleaner.
>> new binary

>>Extra Rust crate or `jq`-manglery in Shell? Neither seems very low friction.

Well, if we convert the compute-pcrs into a library, it removes all those frictions. IMO, a library what we should get from the compute-pcrs repository. It should also pretty straightforward to convert the binary into a library, and then the new binary can rely on this library.

>> update the config maps

>Is API access nearly as easy as what we have from the operator?

Well, the operator would need to fetch the logs and patch the reference values config maps, so why not let the job doing it immediately

Jakob-Naucke:
> > Extra Rust crate or `jq`-manglery in Shell? Neither seems very low friction.
>
> Well, if we convert the compute-pcrs into a library, it removes all those frictions. IMO, a library what we should get from the compute-pcrs repository. It should also pretty straightforward to convert the binary into a library, and then the new binary can rely on this library.

oh yeah that's cleaner indeed. Didn't realize you meant that based on

> include the compute-pcrs binary in a new container image

---

> > Is API access nearly as easy as what we have from the operator?
> 
> Well, the operator would need to fetch the logs and patch the reference values config maps, so why not let the job doing it immediately

What I meant was: The job needs all the k8s API interaction logic again. That's probably fine though.

Jakob-Naucke:
untested, pushed to the wrong remote, bound to happen. converting to draft. sorry for the noise.

compute-pcrs/Containerfile
```diff
@@ -0,0 +1,16 @@
+FROM docker.io/library/rust:1.89-trixie AS builder
+WORKDIR /compute-pcrs
+COPY Cargo.toml Cargo.lock .
+COPY compute-pcrs compute-pcrs
+# Hack: Set compute-pcrs as sole member to avoid needing to copy other crates.
+# In that case, a rebuild would be triggered upon any change in those crates.
+RUN sed -i 's/members =.*/members = ["compute-pcrs"]/' Cargo.toml && \
```
alicefr, line 7: cannot you copy the Cargo.toml from `compute-pcrs` instead?

compute-pcrs/Containerfile
```diff
@@ -0,0 +1,16 @@
...
+    git clone --depth 1 https://github.com/confidential-clusters/reference-values && \
+    cargo build --release -p compute-pcrs
+
+FROM docker.io/library/debian:trixie
+COPY --from=builder /compute-pcrs/target/release/compute-pcrs /usr/local/bin
+COPY --from=builder /compute-pcrs/reference-values /reference-values
+ENTRYPOINT ["compute-pcrs", \
+    "--efivars", "/reference-values/efivars/qemu-ovmf/fcos-42", \
+    "--mokvars", "/reference-values/mok-variables/fcos-42"]
```
alicefr, line 16: nit: usually the args are specified at container creation

compute-pcrs/Containerfile
```diff
@@ -0,0 +1,16 @@
...
+WORKDIR /compute-pcrs
+COPY Cargo.toml Cargo.lock .
+COPY compute-pcrs compute-pcrs
+# Hack: Set compute-pcrs as sole member to avoid needing to copy other crates.
+# In that case, a rebuild would be triggered upon any change in those crates.
+RUN sed -i 's/members =.*/members = ["compute-pcrs"]/' Cargo.toml && \
+    git clone --depth 1 https://github.com/confidential-clusters/reference-values && \
+    cargo build --release -p compute-pcrs
+
+FROM docker.io/library/debian:trixie
```
alicefr, line 11: can we use fedora as base image?

compute-pcrs/src/main.rs
```diff
@@ -0,0 +1,108 @@
...
+    match config_maps
+        .create(&PostParams::default(), &config_map)
+        .await
+    {
+        Ok(_) => info!("Create ConfigMap {}", args.configmap),
+        Err(kube::Error::Api(ae)) if ae.code == 409 => {
+            info!("ConfigMap {} already exists", args.configmap)
+        }
+        Err(e) => return Err(e.into()),
+    }
```
alicefr, line 105: what if the config map already exists? You probably wants to retrieve its value check if it is different from the reference values, and if not then not update it. Right now, it make little sense but when we have more coreos versions to handle then the logic will become useful

manifest-gen/src/main.rs
```diff
@@ -42,6 +43,15 @@ pub struct Args {
...
     #[arg(long, default_value = "operators")]
     trustee_namespace: String,
+
+    #[arg(
+        long,
+        default_value = "quay.io/confidential-clusters/compute-pcrs:latest"
+    )]
+    pcrs_compute_image: String,
+
+    #[arg(long, default_value = "quay.io/fedora/fedora-coreos:42.20250705.3.0")]
```
alicefr, line 53: I would remove this and hardcoded directly in the compute-pcrs. We will then add a logic to detect the coreos version to calculate

operator/src/trustee.rs
```diff
@@ -127,61 +131,77 @@ pub async fn generate_kbs_configurations(
...
+    for (flag, value) in [("configmap", name), ("namespace", trustee_namespace)] {
+        add_flag(flag, value);
+    }
 
-    let config_map = ConfigMap {
-        metadata: kube::api::ObjectMeta {
-            name: Some(name.to_string()),
-            namespace: Some(namespace.to_string()),
+    let pod_spec = PodSpec {
+        service_account_name: Some("cocl-operator".to_string()),
```
alicefr, line 161: I would create a separate service account for the job only with a separate Role to only be able to create and modify the config maps in the trustee namespace

compute-pcrs/Containerfile
alicefr: Kubernetes overwrites the entrypoint in any case. What I don't like is that there is fcos-42 hardcoded there

compute-pcrs/Containerfile
alicefr: Well, it can be built on the debian rust image, but you should copy it in a fedora base image. Or are you afraid that the dynamic library won't match?

manifest-gen/src/main.rs
alicefr: Yes, but the job with the image volume is created by the operator, so you don't need this in the manifests

compute-pcrs/Containerfile
Jakob-Naucke: > Or are you afraid that the dynamic library won't match?

Yes I was, but maybe it's fine. I'll test.

scripts/clean-cluster-kind.sh
```diff
@@ -11,4 +11,5 @@ for image in "$@"; do
 	fi
 done
 kubectl delete deploy cocl-operator -n confidential-clusters || true
+kubectl delete job compute-pcrs -n confidential-clusters || true
```
alicefr, line 14: I think should be handled by the operator

alicefr: If the job completes, the operator can then remove it

operator/src/trustee.rs
```diff
@@ -127,61 +133,82 @@ pub async fn generate_kbs_configurations(
...
-    let create = config_maps
-        .create(&PostParams::default(), &config_map)
-        .await;
-    info_if_exists!(create, "ConfigMap", name);
-
+    let jobs: Api<Job> = Api::namespaced(client.clone(), job_namespace);
+    let create = jobs.create(&PostParams::default(), &job).await;
+    info_if_exists!(create, "Job", job_name);
+    let completed = await_condition(jobs.clone(), job_name, is_job_completed());
+    let _ = timeout(Duration::from_secs(900), completed).await?;
```
alicefr, line 210: Is this blocking until it completes?
The usual way in kubernetes is watching objects with a controller and then trigger a reconciliation loop if there is a change in the state. Can we implement something like this here? See the rust documentation for the [Controller](https://docs.rs/kube/1.1.0/kube/runtime/struct.Controller.html)

compute-pcrs/Containerfile
bgartzi: Has a further decision been taken on this topic? I see this is one of the only references to `reference-values/mok-variables/fcos-42` which I would like to update to `reference-values/mok-variables/fedora-42` for the sake of simplicity.

Are we safe if I move https://github.com/confidential-clusters/reference-values/pull/3 on?

PR #14, opened 2025-08-29
Compute PCRs, using labels and caches

Jakob-Naucke:
- Create new config map to cache known PCR values and parts.
- Fill with PCR label if present in boot image.
- Create new binary as fallback to compute values using an image volume.
- Recompute reference values based on config map.

Supersedes the reference value input workflow and #12.

Uses a config map for value caching instead of a CRD because identifying a CRD with an image ref is a little weird (character type & length limits). I'm not a huge fan either though because this means rewriting the entire map on every new image (_as we do with the reference values…_), so I'm open to other suggestions.

### DIFF
--- a/scripts/clean-cluster-kind.sh
+++ b/scripts/clean-cluster-kind.sh
@@ -4,10 +4,11 @@ set -x
 
 source scripts/common.sh
 
-IMAGE=$1
-if ${RUNTIME} exec -ti kind-control-plane crictl inspecti ${IMAGE} &> /dev/null ; then
-	${RUNTIME} exec -ti kind-control-plane crictl rmi ${IMAGE}
-fi
-
+for image in "$@"; do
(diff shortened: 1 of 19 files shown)


scripts/clean-cluster-kind.sh
```diff
@@ -4,10 +4,12 @@ set -x
...
-fi
-
+for image in "$@"; do
+	if ${RUNTIME} exec -ti kind-control-plane crictl inspecti ${image} &> /dev/null ; then
+		echo "Delete image ${image}"
+		${RUNTIME} exec -ti kind-control-plane crictl rmi ${image}
+	fi
+done
 kubectl delete deploy cocl-operator -n confidential-clusters || true
+kubectl delete job compute-pcrs -n confidential-clusters || true
```
alicefr, line 14: Already left the comment in the other PR, but the job deletion should be handled by the operator

PR #19, opened 2025-09-05
fix: use `matches!` for hex check

yalzhang:
Replace manual char comparison with `matches!` to satisfy clippy.


alicefr:
/lgtm 
@Jakob-Naucke can you please take a look

Issue #1, opened 2025-07-25
Fix trustee installation in a separate namespace


alicefr:
Currently, the trustee installation properly works only if the operator is running the `operators` namespace while we want to keep it in a separate namespace, like `trustee`.

PR #21, opened 2025-09-12
Install the operator and trustee in custom namespaces

alicefr:
Create common variables for the operator and trustee namespace

Fixes: #1


alicefr:
Close this since in #40 we deploy trustee ourself

PR #25, opened 2025-09-16
compute-pcrs/Containerfile: Update container image reference

travier:
We moved the buildroot image into its own repo:
https://github.com/confidential-clusters/buildroot


Jakob-Naucke (approved):
Ah, much cleaner, thanks. FYI I'm also working on moving the operator Containerfile itself to a standardized build container.

PR #29, opened 2025-09-16
github/workflows: Use split workflow instead of matrix

travier:
This split things a bit more and is what we use in other Rust based projects. Let's see if this is useful.

Not sure if that exactly covers what we have already.


travier:
I merged https://github.com/confidential-clusters/cocl-operator/pull/22 but looks like it's not picked up by this PR.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,122 @@
...
+
+# Don't waste job slots on superseded code
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+  # Pinned toolchain for linting
+  ACTIONS_LINTS_TOOLCHAIN: 1.85.0
```
Jakob-Naucke, line 24: this would need to be
```suggestion
  ACTIONS_LINTS_TOOLCHAIN: 1.88.0
```
**or** we relax the MSRV back down to 1.85, #14 removed the constructs that required 1.88

travier: Let's start with 1.85.0 as that's what we use in CoreOS as that's in some version of RHEL. We can bump it back later.

Jakob-Naucke (approved):

PR #33, opened 2025-09-17
cargo: Set MSRV for individual crates

Jakob-Naucke:
#22 set an MSRV for the workspace, but this doesn't have any effect
before setting them for the crates. Set them there. Unblocks #29.


Jakob-Naucke:
was picked into #29, closing

PR #41, opened 2025-09-26
Handle creation of ConfidentialClusters

Jakob-Naucke:
Previous setup had a watcher on cocl objects, but did nothing with it
and just assumed one would be present. Instead, run installations when
=1 cocl is present.

Contains two commits from #36 to make the rebase bearable (sorry)


alicefr:
Looks great!

PR #44, opened 2025-10-02
docs: Add reference values flow

Jakob-Naucke:
Design proposal (partially implemented but open to comments on anything)


docs/design/reference-values.md
```diff
@@ -0,0 +1,95 @@
...
+For better interaction with kubectl, approved images are specified as a very simple custom resource:
+
+```yaml
+apiVersion: confidential-containers.io/v1alpha1
+kind: ApprovedImage
+metadata:
+  name: my-scos
+  namespace: confidential-clusters
+spec:
+  reference: quay.io/my-registry/scos-kernel-layer
```
alicefr, line 31: I'm wondering if we should specifically reference to a an image and its sha here. If we use tags, it can be different images pushed under the same tag. Can this be a problem?

docs/design/reference-values.md
```diff
@@ -0,0 +1,95 @@
...
+Because such explicit updates are assumed to be intended by the cluster administrator, the Confidential Clusters operator can watch the MachineConfigPools and set the images that they reference as approved.
+
+However, kubectl interaction is also supported, both for avoiding reliance on OpenShift and for manual intervention.
+
+## Split data store: CRD for approved images, ConfigMap for PCR parts
+
+For better interaction with kubectl, approved images are specified as a very simple custom resource:
+
+```yaml
+apiVersion: confidential-containers.io/v1alpha1
```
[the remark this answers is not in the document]
Jakob-Naucke: hmm, the API group for `ConfidentialCluster` has been `confidential-containers.io` since [day 1](https://github.com/confidential-clusters/cocl-operator/blame/dc5c59764d0d29dab9eb08b066483e5c335339df/crds/src/lib.rs#L7) but we could move all to `confidential-clusters.io` (or whatever else) because it isn't the same project

docs/design/reference-values.md
Jakob-Naucke: yea this is an issue. Previously, in a comment merged in #14, I postulated
> Non-goal: Support tags whose referenced versions change (e.g. `latest`). This would introduce hard-to-define behavior for disallowing older versions that were introduced as a tag that is still allowed.

which was an easy way of supporting tags for convenience and saying "re-using tags is unsupported, deal with it". Indeed, it would be safer and still relatively user-friendly (especially when using MachineConfig) to support only the SHA. LMK what you think.

alicefr: Yes, I think for now we should support only the shas

docs/design/reference-values.md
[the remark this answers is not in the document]
alicefr: oh yes please, I might have misstype from the beginning then

docs/design/reference-values.md
```diff
@@ -0,0 +1,95 @@
...
+Because a node could be updated again before the second reboot, many combinations of UKI and bootloader components would be considered valid.
+Upon every change of the image PCRs, the reference values that are utilised by Trustee are recomputed with respect to all of these combinations using compute-pcrs.
+A reference value listing for Trustee could then look like this:
+
+```json
+[
+  {
+    "version": "0.1.0",
+    "name": "tpm_pcr4",
+    "expiration": "2026-10-02T13:00:13Z",
```
alicefr, line 84: Is the expiration on the entire list or on each single values?

Jakob-Naucke: It's on the Trustee RVPS resource of the particular name `tpm_pcr4`. As of now, each time the reference values update, we accumulate all and set them to expire in 365 days. This could be refined and then it may (read: did not test) be possible to have multiple reference values with the same name (after all, it's not a dictionary key) but different expiration times.

PR #48, opened 2025-10-07
dependabot: Ignore Warp >=0.4

Jakob-Naucke:
API changed so logging the client address is currently very involved, cf. https://github.com/seanmonstar/warp/issues/1127. Block from updates until resolution.

Alternatively, we could decide logging the client address isn't so important.


alicefr:
We can block for now, if it doesn't resolve we can reevaluate if we can implement our logging eventually.

PR #60, opened 2025-10-16
Separately specify Trustee KBS port and public Trustee address

Jakob-Naucke:
Trustee port was taken from the public address to avoid redundany, but
that made 2.5 assumptions that aren't necessarily true:
- Trustee public address is always known ahead of time.
- Public address will (a) contain a port that is (b) identical to the
  cluster-internal one.

Instead, add the `trusteeKbsPort` field to the cocl CRD. The existing
field `trusteeAddr` is renamed to `publicTrusteeAddr` for clarity.

Drive-by: rename API group as discussed in #44


alicefr:
Thanks Jakob!

PR #61, opened 2025-10-16
Add status field to cocl resource

Jakob-Naucke:
with phases and conditions. Resources are updated as necessary (except for the compute-pcrs image, which is harder to do because it's run by an independent controller).

### DIFF
--- a/crds/src/lib.rs
+++ b/crds/src/lib.rs
@@ -3,9 +3,12 @@
 //
 // SPDX-License-Identifier: MIT
 
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::Condition;
 use kube_derive::CustomResource;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+
+pub mod conditions;
 
 #[derive(CustomResource, Default, Debug, Clone, Deserialize, Serialize, JsonSchema)]
 #[kube(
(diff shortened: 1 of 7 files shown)


operator/src/main.rs
```diff
@@ -29,32 +31,109 @@ mod trustee;
 
 const BOOT_IMAGE: &str = "quay.io/confidential-clusters/fedora-coreos:latest";
 
+fn generate_condition(message: String, reason: String) -> Condition {
+    Condition {
+        last_transition_time: Time(Utc::now()),
+        message,
+        observed_generation: None,
+        reason: reason.clone(),
+        status: "True".to_string(),
```
alicefr: Is it always True?I would still pass as parameter

crds/src/lib.rs
```diff
@@ -28,6 +32,24 @@ pub struct ConfidentialClusterSpec {
...
 
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct ConfidentialClusterStatus {
+    pub conditions: Vec<Condition>,
+}
+
+pub const INSTALLED_CONDITION: &str = "Installed";
+pub const KNOWN_TRUSTEE_ADDRESS_CONDITION: &str = "KnownTrusteeAddress";
+pub const ADDRESS_FOUND_REASON: &str = "AddressFound";
+pub const NO_ADDRESS_FOUND_REASON: &str = "NoAddressFound";
```
alicefr: Can be those be merged into one?

Jakob-Naucke: Whoops, had removed the uses but not the definition. Fixed now.

PR #65, opened 2025-10-27
crds: Port to Golang

Jakob-Naucke:
For API review and potentially more expressive validator logic, port
CRDs and RBAC to kubebuilder (written in Golang).

- Definitions are rendered to YAML, from that to Rust again by kopium
- make: use adaptations of kubebuilder's go-install-tool function to
  install controller-gen (Go) and kopium (Rust)
- `/manifests => /config` because that's kubebuilder's output directory
---
- Based on #61, can be rebased for 4 instead of 8 commits after its merge.
- In a separate commit, contains some RBAC definitions that kubebuilder auto-creates and that I am not sure we should keep.
- RBAC is no longer restricted to a namespace, which I am not sure is supported with kubebuilder.
- @alicefr offline, we discussed porting manifest-gen but I think some parts of it should stay.
- Drive-by: fix empty reference values file, install yq with tooling that this PR introduces

### DIFF
--- a/config/rbac/metrics_auth_role_binding.yaml
+++ b/config/rbac/metrics_auth_role_binding.yaml
@@ -1,0 +1,16 @@
+# SPDX-FileCopyrightText: Jakob Naucke <jnaucke@redhat.com>
+#
+# SPDX-License-Identifier: CC0-1.0
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: metrics-auth-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: metrics-auth-role
+subjects:
+- kind: ServiceAccount
+  name: cocl-operator
+  namespace: system
--- a/config/rbac/confidentialcluster_admin_role.yaml
+++ b/config/rbac/confidentialcluster_admin_role.yaml
@@ -1,0 +1,31 @@
+# SPDX-FileCopyrightText: Jakob Naucke <jnaucke@redhat.com>
+#
+# SPDX-License-Identifier: CC0-1.0
+
+# This rule is not used by the project cocl-operator itself.
+# It is provided to allow the cluster admin to help manage permissions for users.
+#
+# Grants full permissions ('*') over confidential-clusters.io.
+# This role is intended for users authorized to modify roles and bindings within the cluster,
+# enabling them to delegate specific permissions to other users or groups as needed.
+
+apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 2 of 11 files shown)


Jakob-Naucke:
Ready for review now.

operator/src/register_server.rs
```diff
@@ -149,11 +149,9 @@ pub async fn create_register_server_deployment(
         metadata: ObjectMeta {
             name: Some(name.to_string()),
             owner_references: Some(vec![owner_reference]),
-            labels: Some(labels.clone()),
             ..Default::default()
         },
         spec: Some(DeploymentSpec {
-            replicas: Some(1),
```
alicefr: nit: I think this is set by default to one, I would keep it for clarity

api/v1alpha1/machine_types.go
```diff
@@ -0,0 +1,64 @@
+//  SPDX-FileCopyrightText: Alice Frosi <afrosi@redhat.com>
```
alicefr: nit: Any reason why separating the Machine in a second file?

alicefr:
@Jakob-Naucke can you please also commit the manifests?

Nevermind, they are in config now, thanks

crds/src/lib.rs
```diff
@@ -3,43 +3,15 @@
 //
```
alicefr: I would completly removed this file and move all the content in the go definitions

api/v1alpha1/machine_types.go
Jakob-Naucke: The Kubebuilder quickstart suggests this but indeed they can go to one file.

alicefr:
@Jakob-Naucke the PR looks good, and great work! I think in the long run we will need to separate the go part in a new repository but not now. Before merging this, I want to test it manually, I will try this afternoon, or at latest next week

alicefr:
> > I think in the long run we will need to separate the go part in a new repository but not now.
> 
> Why? I'm not a fan of PRs that require lockstep.
> 
Because it is cleaner. Now, we have the mixture of go and rust, I'm not a big fan of. If we separate into a separate repo, we state the our official API is there.

config/rbac/metrics_auth_role_binding.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: metrics-auth-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: metrics-auth-role
+subjects:
+- kind: ServiceAccount
+  name: cocl-operator
+  namespace: system
```
alicefr: Is this namespace somehow overwritten?

PR #68, opened 2025-10-29
Add design document for 2-phase attestation

alicefr:


alicefr:
Outdated, a more recent version will follow

PR #69, opened 2025-11-03
Add unit tests to remainder of operator

Jakob-Naucke:
not including other crates. ~Contains two small non-test fixes.~

### DIFF
--- a/operator/src/main.rs
+++ b/operator/src/main.rs
@@ -199,3 +199,91 @@ async fn main() -> Result<()> {
 
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use chrono::Utc;
+    use cocl_operator_lib::ConfidentialClusterSpec;
+    use http::{Method, Request, StatusCode};
+    use k8s_openapi::apimachinery::pkg::apis::meta::v1::Time;
+    use kube::{api::ObjectList, client::Body};
+
+    use super::*;
+    use crate::mock_client::*;
+
+    fn dummy_cocl() -> ConfidentialCluster {
+        ConfidentialCluster {
+            metadata: ObjectMeta {
+                name: Some("test".to_string()),
+                ..Default::default()
+            },
+            status: None,
+            spec: ConfidentialClusterSpec {
+                trustee_image: "".to_string(),
+                pcrs_compute_image: "".to_string(),
+                register_server_image: "".to_string(),
+                public_trustee_addr: None,
+                register_server_port: None,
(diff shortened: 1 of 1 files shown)


operator/src/reference_values.rs
```diff
@@ -294,4 +298,78 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn test_job_reconcile_no_name() {
+        let clos = |req: &Request<_>| panic!("unexpected API interaction: {req:?}");
+        let ctx = generate_rv_ctx(MockClient::new(clos, "test".to_string()).into_client());
+        let job = Arc::new(Default::default());
+        let err = job_reconcile(job, Arc::new(ctx)).await.err().unwrap();
+        assert!(err.to_string().contains("but had no name"));
+    }
```
alicefr: Can this happen?

operator/src/reference_values.rs
```diff
@@ -294,4 +298,78 @@ mod tests {
...
+        let clos = |req: &Request<_>| panic!("unexpected API interaction: {req:?}");
+        let ctx = generate_rv_ctx(MockClient::new(clos, "test".to_string()).into_client());
+        let mut job = dummy_job();
+        job.status = None;
+        let err = job_reconcile(Arc::new(job), Arc::new(ctx))
+            .await
+            .err()
+            .unwrap();
+        assert!(err.to_string().contains("but had no status"));
+    }
```
alicefr: I'm also not sure about this test as well, isn't kubernetes that set the job status?

operator/src/reference_values.rs
alicefr: Yes, I would drop both

operator/src/main.rs
```diff
@@ -287,4 +291,96 @@ mod tests {
...
+            r if [Method::GET, Method::PATCH].contains(r.method()) => {
+                Ok(serde_json::to_string(&dummy_cocl()).unwrap())
+            }
+            _ => panic!("unexpected API interaction: {req:?}"),
+        };
+        let client = MockClient::new(clos, "test".to_string()).into_client();
+        let mut cocl = dummy_cocl();
+        cocl.metadata.deletion_timestamp = Some(Time(Utc::now()));
+        let result = reconcile(Arc::new(cocl), Arc::new(client)).await.unwrap();
+        assert_eq!(result, Action::await_change());
```
alicefr: Shouldn't we expect the condition for the uninstalling?

PR #77, opened 2025-11-05
build: Generate Rust CRDs inside build container

Jakob-Naucke:
Were previously built on the host, but this makes unnecessary assumptions on the build process and host. Build using Makefile.

Unify build directories across Containerfiles for better cache hits.

---
Draft but already requesting review @alicefr because we should decide before buildroot is updated. Buildroot will need to be updated anyhow, this approach additionally installs **golang** and **rustfmt**.

In this draft, the CRDs are built inside the container using the Makefile.
This approach is simpler and requires less double-tracking between Makefile and Containerfile if the setup changes. It's a little less clean because it hides what's happening from the Containerfile reader.

The best alternative I can think of runs everything from the Containerfile:
- Also install yq in the buildroot.
- Create a file (TOML or whatever) to track tool versions that we use, of controller-gen and kopium. A bonus is that kopium could be set as dev dependency in Cargo.toml and then be picked up by dependabot (ignoring that we'll lock it anyhow). I'm not aware of such a simple approach for updating a Go-based tool dependency though.
- Have Containerfile & Makefile read from this versions file to determine and install tools.

PS: If you find installing rustfmt to the build container too ugly, we can also move the formatting of kopium-generated files elsewhere. Doing so makes the lint CI slightly weirder, but is totally possible.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -7,10 +7,13 @@ ARG build_type
 # Dependency build stage
 FROM ghcr.io/confidential-clusters/buildroot AS builder
 ARG build_type
-WORKDIR /cocl-operator
+WORKDIR /build
 
-COPY Cargo.toml Cargo.lock .
+COPY Cargo.toml Cargo.lock go.mod go.sum Makefile .
+COPY api api
 COPY lib lib
+RUN make crds-rs
+
 COPY operator/Cargo.toml operator/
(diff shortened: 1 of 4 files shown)


alicefr:
/lgtm 
Since now the build in Konflux is successful, see https://github.com/joelcapitao/cocl-operator/pull/1

PR #80, opened 2025-11-11
Make deployment cluster spec fields immutable

Jakob-Naucke:
Updates are currently supported but reconciliation is improperly
designed as it does not let the status follow. Make images & ports
immutable.

and a lot of other small fixes


Jakob-Naucke:
> Looks much cleaner, thanks!

is that approval 🙃?

Jakob-Naucke:
resolved a permissions issue that was indeed uncovered in integration testing :)

PR #101, opened 2025-11-19
feat(operator): Add initial OLM bundle for the operator

yalzhang:
Introduces OLM bundle generation for the operator, enabling OLM-based
deployment and management.

Key changes include:
- `Makefile` targets (`bundle`, `bundle-image`, `push-bundle`, `push-all`)
  for automated bundle generation, building, and pushing.
- New `bundle/` directory structure for OLM artifacts, with static templates
  and generated manifests.
- Security enhancement in `Containerfile` to run as a non-root user.
- Updated `README.md` with OLM bundle management workflows.
- `bundle/manifests/` added to `.gitignore`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,6 +101,31 @@ push: image
 release-tarball: manifests
 	tar -cf trusted-execution-operator-$(TAG).tar config
 
+# OLM Bundle related variables
+BUNDLE_DIR := bundle
+BUNDLE_IMAGE := $(REGISTRY)/trusted-cluster-operator-bundle:$(TAG)
+BUNDLE_PACKAGE ?= trusted-cluster-operator
+PREVIOUS_CSV ?= ""  # optional previous CSV for OLM upgrades
+
+.PHONY: bundle bundle-image push-bundle
+
+bundle: manifests
+	@echo "Generating OLM bundle..."
+	@OPERATOR_IMAGE=$(OPERATOR_IMAGE) \
+	COMPUTE_PCRS_IMAGE=$(COMPUTE_PCRS_IMAGE) \
+	REG_SERVER_IMAGE=$(REG_SERVER_IMAGE) \
+	scripts/generate-bundle-prod.sh -v $(TAG) $(if $(PREVIOUS_CSV),-p $(PREVIOUS_CSV))
+
+bundle-image: bundle
+	@echo "Building OLM bundle image..."
+	$(CONTAINER_CLI) build -f $(BUNDLE_DIR)/Containerfile -t $(BUNDLE_IMAGE) $(BUNDLE_DIR)/
+
+push-bundle: bundle-image
+	@echo "Pushing OLM bundle image..."
+	$(CONTAINER_CLI) push $(BUNDLE_IMAGE) $(PUSH_FLAGS)
+
+push-all: push push-bundle ## Pushes all operator and bundle images
+
 install: $(YQ)
 ifndef TRUSTEE_ADDR
 	$(error TRUSTEE_ADDR is undefined)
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -82,20 +82,42 @@ cluster-cleanup:
...
+	mkdir -p bundle/manifests bundle/metadata
+	cp config/crd/*.yaml bundle/manifests/
+	cp bundle/static/metadata/annotations.yaml bundle/metadata/
+	cp bundle/static/Containerfile bundle/
+	$(YQ) '.metadata.name = "trusted-cluster-operator.v$(TAG)"' bundle/static/manifests/trusted-cluster-operator.clusterserviceversion.yaml | \
+	$(YQ) '.spec.version = "$(TAG)"' | \
+	$(YQ) '.metadata.annotations.containerImage = "$(OPERATOR_IMAGE)"' | \
+	$(YQ) '.spec.install.spec.deployments[0].spec.template.spec.containers[0].image = "$(OPERATOR_IMAGE)"' | \
+	$(YQ) '(.spec.install.spec.deployments[0].spec.template.spec.containers[0].env[] | select(.name == "COMPUTE_PCRS_IMAGE")).value = "$(COMPUTE_PCRS_IMAGE)"' | \
+	$(YQ) '(.spec.install.spec.deployments[0].spec.template.spec.containers[0].env[] | select(.name == "REGISTER_SERVER_IMAGE")).value = "$(REG_SERVER_IMAGE)"' \
```
Jakob-Naucke: nit:
```suggestion
	$(YQ) '.metadata.name = "trusted-cluster-operator.v$(TAG)" |
		.spec.version = "$(TAG)" |
		.metadata.annotations.containerImage = "$(OPERATOR_IMAGE)" |
		.spec.install.spec.deployments[0].spec.template.spec.containers[0].image = "$(OPERATOR_IMAGE)" |
		(.spec.install.spec.deployments[0].spec.template.spec.containers[0].env[] | select(.name == "COMPUTE_PCRS_IMAGE")).value = "$(COMPUTE_PCRS_IMAGE)" |
		(.spec.install.spec.deployments[0].spec.template.spec.containers[0].env[] | select(.name == "REGISTER_SERVER_IMAGE")).value = "$(REG_SERVER_IMAGE)"' \
		bundle/static/manifests/trusted-cluster-operator.clusterserviceversion.yaml \
```

Jakob-Naucke (commented):
Thanks @yalzhang! Now that there is a target to generate the manifests, I wonder if the dynamic ones should just be gitignored and treated as "object code", like the Rust CRD translations. WDYT?

config/rbac/role_binding.yaml
```diff
@@ -16,4 +16,3 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: trusted-cluster-operator
-  namespace: system
```
yalzhang: @Jakob-Naucke  I feel like I shouldn’t update these files since they’re generated by Kubebuilder, but hardcoding the namespace here doesn’t seem right either. What’s your take? If that’s the case, maybe we should fix it at the source—perhaps with Kustomize patches? I’m not sure if modifying api/trusted-cluster-gen.go would be too heavy-handed.

Jakob-Naucke: Indeed, this is a placeholder filled by kustomize, so currently the flow is something like
- edit namespace in `config/rbac/kustomization.yaml`
- `kubectl apply -k config/rbac`
If you find this overly hardcoded, we could also consider a template.

bundle/static/manifests/trusted-cluster-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,139 @@
+---
```
alicefr: This file seems generated, do we need to commit it?

alicefr (approved):

Jakob-Naucke (approved):

PR #102, opened 2025-11-19
Add ApprovedImage resource 

Jakob-Naucke:
Allow administrators to add and remove approved images with a new CRD.

- Add optional generation of such a resource to cocl-gen + Makefile.
- Use finalizer to handle deletion.


Jakob-Naucke (commented):
discussed offline, note to self: editing doesn't make much sense because two must be approved concurrently anyhow. make immutable and remove editor role.

PR #116, opened 2025-12-22
trustee: Configure reference values based on all possible PCR combinations

bgartzi:
Now the compute-pcrs library supports computing PCR value combinations. In other words, given all events from image A and image B (where some of the components got updated), it computes all possible PCRs of the intermediate states that a node could go through during an update from image A to image B.

While implementing this I found a compute-pcrs bug that this PR relies on: https://github.com/trusted-execution-clusters/compute-pcrs/pull/56

For now, I added a workaround commit that is not signed-off-by me. Once the compute-pcrs library gets merged, I will update this PR.

Another discussion topic that comes to my mind is how unit tests are implemented to face this new integration. They sure could be way more isolated from compute-pcrs, which I will update if relevant.


Jakob-Naucke (commented):
Very nice. A few comments.

attestation-key-register/Containerfile
```diff
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: CC0-1.0
 
 ARG build_type
-FROM ghcr.io/trusted-execution-clusters/buildroot AS builder
+FROM ghcr.io/trusted-execution-clusters/buildroot:fedora AS builder
```
Jakob-Naucke: `:latest` and `:fedora` point to Fedora 43 images that are apparently different, but I don't even know how they're different. Is there a difference that matters to this PR?

Jakob-Naucke (approved):

Jakob-Naucke (approved):

bgartzi:
Rebased the branch. Felt free to add some changes to the pcr configmap verification logic to simplify it a bit.

Integration tests are broken now due to https://github.com/trusted-execution-clusters/operator/issues/307 and https://github.com/trusted-execution-clusters/reference-values/pull/8

bgartzi:
Without proper warmup, `test_combined_image_pcrs_configmap_updates` is still a bit flaky due to the 180s timeout in `verify_expected_pcrs`. I could just increase it to 300 and call it good, but I might end up adding the secondary approved image to the images that are pre-pulled to the cluster, which might be a better thing to do.

operator/src/trustee.rs
```diff
@@ -72,22 +74,28 @@ pub fn get_image_pcrs(image_pcrs_map: ConfigMap) -> Result<ImagePcrs> {
...
+        .map(|v| v.pcrs.iter().flat_map(|p| p.events.clone()).collect())
+        .collect();
+
+    let pcr_combinations = combine_images(&tpm_events);
+    for pcr in pcr_combinations.iter().flatten() {
         reference_values_in
             .entry(format!("pcr{}", pcr.id))
             .or_default()
-            .push(JsonString(pcr.value.clone()));
+            .insert(hex::encode(pcr.value.clone()));
```
Jakob-Naucke: nit: no need to clone `hex::encode` args, it takes `AsRef` (also two in `get_image_pcrs_success`)

Jakob-Naucke (commented):
> Without proper warmup, `test_combined_image_pcrs_configmap_updates` is still a bit flaky due to the 180s timeout in `verify_expected_pcrs`. I could just increase it to 300 and call it good, but I might end up adding the secondary approved image to the images that are pre-pulled to the cluster, which might be a better thing to do.

I agree this is nicer because it decouples timeouts from the test host's internet speed. I noticed we don't even pre-pull in CI at all yet. Could you add it to `.github/workflows/integration-tests.yml` before the integration tests?

bgartzi:
Now extra approved images can be also pre-pulled, defaults to the secondary approved image pulled in the combination integration test and integration tests in github pre-pull images first.

scripts/pre-pull-images.sh
```diff
@@ -8,11 +8,25 @@
...
+	docker tag "$image" "$local_image"
+	docker push "$local_image"
+	docker exec -ti kind-control-plane crictl pull "$local_image"
+	echo "-------------------------------"
+}
+
+index=0
+pull_approved_image "$APPROVED_IMAGE" $index
+
+for image in $EXTRA_APPROVED_IMAGES; do
```
Jakob-Naucke: I'm not a fan of the extra env. I'd rather keep these out of the Makefile, put a const for the test image at the top of `tests/trusted_execution_cluster.rs`, and grep-sed for that here, WDYT?


```suggestion
combined_pcrs_test_image=$(grep "const COMBINED_PCRS_TEST_IMAGE" tests/trusted_execution_cluster.rs | \
                           sed -E 's/.*"(.*)";/\1/')
for image in "$APPROVED_IMAGE" "$combined_pcrs_test_image"; do
```

Jakob-Naucke (commented):
urgh, two start failures in the test (probably not your fault) and a stop failure (fix is coming soon). I'll try and improve these soon.

scripts/pre-pull-images.sh
bgartzi: skipped the for, but I can put it there. You tell me.

Jakob-Naucke (approved):
no, I like it. @alicefr PTAL too bc it's pretty big

api/trusted-cluster-gen.go
```diff
@@ -184,34 +195,33 @@ func generateTrustedExecutionClusterCR(args *Args) error {
...
-		return nil
-	}
+	for i, approvedImage := range args.approvedImages {
+		approvedImage := &v1alpha1.ApprovedImage{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: v1alpha1.GroupVersion.String(),
+				Kind:       "ApprovedImage",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      fmt.Sprintf("coreos-%d", i),
```
alicefr: can we use something different then the index? maybe the instead the approvedImage with the `/` and `:`replaced by `-`. Or do we have a limit in the name lenght?

alicefr: Also `coreos` wasn't great by the way

test_utils/src/constants.rs
```diff
@@ -0,0 +1,233 @@
...
+pub const TEC_NAME: &str = "trusted-execution-cluster";
+pub const APPROVED_IMAGE_NAME: &str = "coreos-0";
+pub const TRUSTEE_CONFIG_MAP: &str = "trustee-data";
+pub const RV_JSON_KEY: &str = "reference-values.json";
+
+#[macro_export]
+macro_rules! expected_pcr7_hash {
+    () => {{ "b3a56a06c03a65277d0a787fcabc1e293eaa5d6dd79398f2dda741f7b874c65d" }};
+}
+
```
alicefr: One question, do we expect these constant to remain always the same or do we need to modify them for example when we upgrade the fedora version of the approved image?

tests/trusted_execution_cluster.rs
```diff
@@ -353,3 +353,69 @@ async fn test_approved_image_readoption() -> anyhow::Result<()> {
     Ok(())
 }
 }
+
+named_test! {
+async fn test_combined_image_pcrs_configmap_updates() -> anyhow::Result<()> {
+    let test_ctx = setup!([
+        "quay.io/trusted-execution-clusters/fedora-coreos@sha256:372a5db90a8695fafc2869d438bacd7f0ef7fd84f63746a450bfcd4b8b64ae83",
```
alicefr: Question, how do we upgrade this? Should it make sense to move the definition out of the test and put an env variable. Same question, how do we get the expected pcr value.

tests/trusted_execution_cluster.rs
```diff
@@ -22,6 +22,8 @@ use trusted_cluster_operator_lib::{
 use trusted_cluster_operator_test_utils::constants::*;
 use trusted_cluster_operator_test_utils::*;
 
+const COMBINE_PCRS_UPDATE_TEST_IMAGE: &str = "quay.io/trusted-execution-clusters/fedora-coreos@sha256:372a5db90a8695fafc2869d438bacd7f0ef7fd84f63746a450bfcd4b8b64ae83";
```
alicefr: This is make sense to squash the definition of this image in the commit where this was introduced

alicefr:
@bgartzi I did a first pass, the only part that worries me is how we update the various hardcoded pcrs values and the second image when we upgrade the whole setup

tests/trusted_execution_cluster.rs
Jakob-Naucke: I mean, it's just a second version for testing that combination works, we can probably leave it for a few years. It's not a moving target like FCOS versions.

test_utils/src/constants.rs
bgartzi: They might change with approved images, specially pcr4. If we'd like to parametrize this so we could hardcode this upper in the stack, such as in the makefile, the problem wouldn't be passing pcr values, but the events of each of them.

tests/trusted_execution_cluster.rs
bgartzi: As soon as we want to write another test that handles 2 approved images and their upgrade path, yes. We can get ahead of time and put it that way if you want. I don't have a personal choice.

When it comes to upgrades, if we would like to change the image to a completely new one, we need to boot a VM with that image, check the tpm event log and look for shim, grub and kernel related events. Then we need to compute possible combinations, which include primary bootloader + secondary kernel and secondary bootloader + primary kernel.

tests/trusted_execution_cluster.rs
bgartzi: Did you also mean the pre-pull logic? Just the const definition, right?

bgartzi:
still struggling to remember I have to sign commits sorry.

@alicefr:
> @bgartzi I did a first pass, the only part that worries me is how we update the various hardcoded pcrs values and the second image when we upgrade the whole setup

The problem is we could automate this using the compute-pcrs library somewhere in CI or makefile, but that would somehow break the purpose of the end to end test. If we really wish to test the whole stack, we need to boot a VM with the tested images, extract all pcr events and hashes, then update the values. We could automate it, but it is not precisely trivial.

tests/trusted_execution_cluster.rs
alicefr: Just the const definition

tests/trusted_execution_cluster.rs
alicefr: can we document in a comment how we build this image and describe how we extracted the pcrs values. At least, we won't forget it

api/trusted-cluster-gen.go
bgartzi: Wrote a some tweaks on trusted-cluster-gen and things that use it to allow configuring approved image references together with their names.
Then I tried giving approved images more specific names.

test_utils/src/constants.rs
bgartzi: Got rid of unnecessary macros and put those values into constants with more clear names. Then, yes, I keep using some macros as helpers to build tpmevents out of constant values that should be now easier to update.

tests/trusted_execution_cluster.rs
bgartzi: Added a document in `docs/dev/test_approved_images.md` trying to explain what each image carry, and some insights on how to calculate the pcrs, and combined pcrs upon the event of an approved image update.

bgartzi:
Now, I believe this is ready for another look @alicefr  unless something in CI fails. @Jakob-Naucke you're also invited ofc.

bgartzi:
Of course I moved some constant around and changed their name but forgot to update their reference in the document that covered them. Fixed now.

bgartzi:
Last failure was about kubevirt installation timing out. Trying again.

alicefr (approved):

alicefr:
@bgartzi sorry for the delay. It looks great thanks for the work

bgartzi:
np :)

Jakob-Naucke:
@dependabot unignore compute-pcrs-lib

PR #154, opened 2026-01-14
Add design for the bootstrap phase

alicefr:


patrickdillon (commented):
Can we identify what details (at a high-level) the user will need to pass to the installer? Perhaps just the address of the external attestation server and key?

alicefr:
Close this. a more updated version will follow

PR #172, opened 2026-01-29
Update Container images to fedora 43

yairpod:
Updating all Fedroa container images to version 43. Also updating rust version to 1.92

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -22,7 +22,7 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   # Pinned toolchain for linting
-  ACTIONS_LINTS_TOOLCHAIN: 1.85.0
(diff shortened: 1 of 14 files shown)


alicefr (approved):

Jakob-Naucke (commented):
LGTM, but as per the lint result, you'll also need to update the toolchain in `.github`. See also #116.

> But I more wondering if we should tag the buildroot with the fedora version instead of latest.

@alicefr yes, `:fedora` is whatever buildroot updates to, and incidentally, we've been building on an F43 (not F42) buildroot that's slowly been running out of date. We could move buildroot to tagging the version too (and then, ideally, having dependabot update to latest not rawhide)

Jakob-Naucke (approved):

alicefr (approved):

PR #181, opened 2026-02-04
Adding quick fail if the virtctl command is missing

yairpod:
Tests depending on the virtctl command will now fail before starting VMs and will not wait for the timeout. Instead they will report the error clearly.


Jakob-Naucke (commented):
I'd dare remove this check from `virtctl_ssh_exec` now

PR #237, opened 2026-04-09
Quit RV controllers upon TEC recreation

Jakob-Naucke:
to prevent racing of controllers on server side apply

Fixes: #216

@alicefr quirks discovered, lmk if they should be bugs:
- creating ApprovedImage without a TEC present will result in the image silently not being created
- deleting TEC without an ApprovedImage's RVs computed will leave the ApprovedImage behind, not garbage collected as #187 perhaps intended?

### DIFF
--- a/operator/src/main.rs
+++ b/operator/src/main.rs
@@ -47,6 +47,8 @@ struct ClusterContext {
     client: Client,
     /// UID of cluster that watchers are based on
     uid: Mutex<Option<String>>,
+    /// Handles for reference value controller tasks, aborted on TEC recreation
+    rv_handles: Mutex<Vec<tokio::task::JoinHandle<()>>>,
 }
 
 fn is_installed(status: Option<TrustedExecutionClusterStatus>) -> bool {
@@ -77,6 +79,12 @@ async fn launch_rv_watchers(
         warn!("Failed to acquire lock on context UID store");
     }
     if launch_watchers {
+        // Abort any previously spawned controllers before launching new ones
+        if let Ok(mut handles) = ctx.rv_handles.lock() {
+            for handle in handles.drain(..) {
+                handle.abort();
+            }
+        }
         info!(
             "First registration of TrustedExecutionCluster {name} by this operator. \
              Launching reference value watchers."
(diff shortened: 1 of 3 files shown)


alicefr:
> to prevent racing of controllers on server side apply
> 
> Fixes: #216
> 
> @alicefr quirks discovered, lmk if they should be bugs:
> 
> * creating ApprovedImage without a TEC present will result in the image silently not being created

Approved images are a bit independent from TEC,  so I think they could be created head
> * deleting TEC without an ApprovedImage's RVs computed will leave the ApprovedImage behind, not garbage collected as [Add missing ownership to approved images, machine and AK #187](https://github.com/trusted-execution-clusters/operator/pull/187) perhaps intended?

We can debate a bit on this. But I think ApprovedImages can be defined by the cluster admin manually or by another operator so they shouldn't be cleaned up by our operator. Although, in the test probably they remain

alicefr:
I think we need to review if the approved image should be owned by TEC. I personally don't think so anymore. Then, I don't think this solution is required

Jakob-Naucke:
> * creating ApprovedImage without a TEC present will result in the image silently not being created

is a quirk of this fix, not an existing one because the watcher should quit on TEC deletion not on recreation. converting to draft.

> * deleting TEC without an ApprovedImage's RVs computed will leave the ApprovedImage behind, not garbage collected as #187 perhaps intended?

~cannot reproduce~ correction: when deleting TEC before operator was running, the ApprovedImage will not have been adopted yet. this is probably expected.

Jakob-Naucke:
Updated, back out of draft. Watcher should _not_ quit on TEC deletion because it won't be able to remove the finalizer, instead, do not attempt computation on addition if no TEC present. @alicefr PTAL or delay that look until after CI revival if you prefer

alicefr:
> Updated, back out of draft. Watcher should _not_ quit on TEC deletion because it won't be able to remove the finalizer, instead, do not attempt computation on addition if no TEC present. @alicefr PTAL or delay that look until after CI revival if you prefer

@Jakob-Naucke why do we need TEC for the computation? Is because the image of compute PCR is there?

Jakob-Naucke:
Indeed, the image-pcrs configmap is owned by the TEC object, so the computed PCRs have nowhere to go without a TEC present

alicefr:
Then maybe we could prioritize this https://github.com/trusted-execution-clusters/operator/issues/215 , and make approved images independent from the TEC CR

UPDATE: what if we just create the configmap independently of TEC and then let adopted by TEC CR when created

alicefr:
In general, I think it is easier to react when objects are created rather the retroactively. Otherwise, we need a way to understand when the approved images existed but they weren't calculated yet

Jakob-Naucke:
Closing in favour of #246

PR #240, opened 2026-04-14
Create webhook to prevent more than one TEC

Jakob-Naucke:
more powerful than just giving a NonUnique status

Fixes: https://github.com/trusted-execution-clusters/operator/issues/225

@alicefr I wrote this with a validating webhook as suggested by #225 -- afaict this is what you need if you're going to have one, but I'm not happy with the design at all. Should we introduce a single permitted name instead?


alicefr:
> Should we introduce a single permitted name instead?

I really would like to avoid that since it is too restrictive. Might also not work with downstream if we want to name it confidential clusters

alicefr:
The alternative is that we don't error out at creation but we set in the second CR status that it is an invalid configuration

alicefr:
we haven't design yet, but we will need to design the upgrade path for the operator and the crd. Not sure if in that case we will have 2 TEC CR

alicefr:
The webhook seems the strogest option to prevent the user to install 2 configuration. What don't you like?

Jakob-Naucke:
> > Should we introduce a single permitted name instead?
> 
> I really would like to avoid that since it is too restrictive. Might also not work with downstream if we want to name it confidential clusters

We could rename in downstream. FWIW this is an approach e.g. for many OpenShift resources, like [Proxy](https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/configuring_network_settings/enable-cluster-wide-proxy#nw-proxy-configure-object_config-cluster-wide-proxy)

> The alternative is that we don't error out at creation but we set in the second CR status that it is an invalid configuration

We do that, #225 asserted it wasn't enough but maybe it is.

> The webhook seems the strogest option to prevent the user to install 2 configuration. What don't you like?

Amount of boilerplate, certificate management, ports used, not atomic

alicefr:
> > > Should we introduce a single permitted name instead?
> > 
> > 
> > I really would like to avoid that since it is too restrictive. Might also not work with downstream if we want to name it confidential clusters
> 
> We could rename in downstream. FWIW this is an approach e.g. for many OpenShift resources, like [Proxy](https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/configuring_network_settings/enable-cluster-wide-proxy#nw-proxy-configure-object_config-cluster-wide-proxy)
> 

Ok if this is a common pattern, then I'm ok with it. We can overwrite the name in the bundle in the confidental cluster repository

> > The alternative is that we don't error out at creation but we set in the second CR status that it is an invalid configuration
> 
> We do that, #225 asserted it wasn't enough but maybe it is.
> 
> > The webhook seems the strogest option to prevent the user to install 2 configuration. What don't you like?
> 
> Amount of boilerplate, certificate management, ports used, not atomic

Agree

Jakob-Naucke:
Closing for now because
> we haven't design yet, but we will need to design the upgrade path for the operator and the crd. Not sure if in that case we will have 2 TEC CR

is unresolved

PR #196, opened 2026-02-13
Add cert-manager support

Jakob-Naucke:
cc @alicefr. Haven't written docs yet but you can get an idea of where it's going by looking at `api/trusted-cluster-gen.go`, `test_utils::set_certificates`.

Doesn't include verification for attestation key register. I used the same method as for register server, but I was under the impression the root certificate added in Ignition, which works for the merge from register server, isn't loaded at that time?
```
HTTP request failed: Put "https://attestation-key-register.test-bc1e06e7.svc.cluster.local:8001/register-ak": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

Didn't bother too much because that flow is subject to updates.

Fixes: #2
(at which I took first looks when it was new, hence the longest commit timespan in a PR from me to date)

PR #242, opened 2026-04-16
Implement KBS API integration for LUKS key and AK management 

iroykaufman:
Currently, every time the LUKS key or the AK is updated, the operator patches the trustee deployment, which causes a restart of the pod. This PR introduces a way to avoid this by setting the AK and LUKS using the KBS API. 

## Core implementation points:
* Right before the trustee first deploys the operator, create an auth key for the KBS API and save it as a kube secret.
* Replacing the patch mechanism with API calls
* Add a reconcile loop that watches for changes in the trustee deployment and sync the AK and LUKS key that was lost.

## Tests:
* test_luks_key_sync - this test checks that the luks key is first sent to tustee and validates that after trustee restarts, the keys are sent again.
* test_attestation_key_sync - The same flow as with the luks key test.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -23,6 +23,8 @@ RUN sed -i 's/members = .*/members = ["lib", "operator"]/' Cargo.toml && \
 # In debug builds, build dependencies to avoid full rebuild.
 RUN if [ "$build_type" = debug ]; then cargo build -p operator; fi
 
+RUN dnf install -y perl-FindBin perl-core
+
 # Target build stage
 COPY operator/src operator/src
 RUN cargo build -p operator $(if [ "$build_type" = release ]; then echo --release; fi)
--- a/operator/Cargo.toml
+++ b/operator/Cargo.toml
@@ -30,7 +30,7 @@ serde.workspace = true
 serde_json.workspace = true
 thiserror = "2.0.18"
 tokio.workspace = true
-kbs-client = {git = "https://github.com/confidential-containers/trustee"}
+kbs-client = {git = "https://github.com/iroykaufman/trustee/", branch = "ak-registration"}
 
 [dev-dependencies]
 http.workspace = true
--- a/operator/Cargo.toml
+++ b/operator/Cargo.toml
@@ -30,6 +30,7 @@ serde.workspace = true
 serde_json.workspace = true
 thiserror = "2.0.18"
 tokio.workspace = true
+kbs-client = {git = "https://github.com/confidential-containers/trustee"}
 
 [dev-dependencies]
 http.workspace = true
--- a/operator/src/kbs-config.toml
+++ b/operator/src/kbs-config.toml
@@ -4,6 +4,7 @@ insecure_http = true
 
 [admin]
 insecure_api = true
+auth_public_key="/key/public.pub"
 type = "InsecureAllowAll"
 
 [attestation_token]
--- a/operator/src/kbs-config.toml
+++ b/operator/src/kbs-config.toml
@@ -33,8 +33,7 @@ policy_engine = "opa"
 
 [[plugins]]
 name = "resource"
-type = "LocalFs"
-dir_path = "/opt/trustee/kbs-repository"
+type = "kvstorage"
 
 [policy_engine]
 policy_path = "/opt/trustee/policy.rego"
--- a/operator/src/main.rs
+++ b/operator/src/main.rs
@@ -179,6 +179,11 @@ async fn install_trustee_configuration(
         Err(e) => error!("Failed to create the attestation policy configmap: {e}"),
     }
 
+    match trustee::generate_trustee_auth_keys_secret(client.clone(), owner_reference.clone()).await {
+        Ok(_) => info!("Generate auth keys for the KBS API",),
+        Err(e) => error!("Failed to create the auth keys: {e}"),
+    }
+
     let kbs_port = cluster.spec.trustee_kbs_port;
     match trustee::generate_kbs_service(client.clone(), owner_reference.clone(), kbs_port).await {
         Ok(_) => info!("Generate the KBS service"),
--- a/operator/src/main.rs
+++ b/operator/src/main.rs
@@ -10,6 +10,7 @@ use std::time::Duration;
 use anyhow::Result;
 use env_logger::Env;
 use futures_util::StreamExt;
+use k8s_openapi::api::apps::v1::Deployment;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::Condition;
 use kube::runtime::controller::{Action, Controller};
 use kube::runtime::watcher;
@@ -17,6 +18,7 @@ use kube::{Api, Client};
 use log::{error, info, warn};
 
 use operator::generate_owner_reference;
+use trusted_cluster_operator_lib::endpoints::TRUSTEE_DEPLOYMENT;
 use trusted_cluster_operator_lib::{TrustedExecutionCluster, TrustedExecutionClusterStatus};
 use trusted_cluster_operator_lib::{conditions::*, update_status};
 
@@ -195,9 +197,22 @@ async fn install_trustee_configuration(
         "RELATED_IMAGE_TRUSTEE",
         "quay.io/trusted-execution-clusters/key-broker-service:20260106",
     );
-    match trustee::generate_kbs_deployment(client, owner_reference, &trustee_image).await {
+    match trustee::generate_kbs_deployment(client.clone(), owner_reference, &trustee_image).await {
         Ok(_) => info!("Generate the KBS deployment"),
         Err(e) => error!("Failed to create the KBS deployment: {e}"),
+    }
+
+    // Wait for the KBS pod to be ready before syncing secrets
+    let deployments: Api<Deployment> =
+        Api::default_namespaced(client.clone());
+    match trustee::wait_for_deployment_ready(&deployments).await {
+        Ok(_) => {}
+        Err(e) => error!("Error waiting for {TRUSTEE_DEPLOYMENT} to be ready: {e}"),
+    }
+
+    match trustee::sync_all_machine_secrets(client).await {
+        Ok(_) => info!("Synced all machine secrets to KBS"),
+        Err(e) => error!("Failed to sync machine secrets to KBS: {e}"),
     }
 
     Ok(())
--- a/operator/src/main.rs
+++ b/operator/src/main.rs
@@ -10,7 +10,6 @@ use std::time::Duration;
 use anyhow::Result;
 use env_logger::Env;
 use futures_util::StreamExt;
-use k8s_openapi::api::apps::v1::Deployment;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::Condition;
 use kube::runtime::controller::{Action, Controller};
 use kube::runtime::watcher;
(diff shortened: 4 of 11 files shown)


Containerfile
```diff
@@ -23,6 +23,8 @@ RUN sed -i 's/members = .*/members = ["lib", "operator"]/' Cargo.toml && \
 # In debug builds, build dependencies to avoid full rebuild.
 RUN if [ "$build_type" = debug ]; then cargo build -p operator; fi
 
+RUN dnf install -y perl-FindBin perl-core
```
alicefr, line 26: Why do we need these dependecies?

operator/src/trustee.rs
```diff
@@ -156,6 +157,31 @@ fn generate_secret_volume(id: &str) -> (Volume, VolumeMount) {
...
+        .context("Auth private key is not valid UTF-8")?;
+
+    let secret = secret_api.get(id).await?;
+    let secret_data = secret.data.context("Secret has no data")?;
+    let resource_bytes = secret_data
+        .get("root")
+        .context("Secret missing root key")?
+        .0
+        .clone();
+    let url = format!("http://{TRUSTEE_SERVICE}:{TRUSTEE_PORT}");
```
alicefr, line 163: We definitely https here, when https://github.com/trusted-execution-clusters/operator/pull/196 is merged, you should test it with it

operator/src/trustee.rs
```diff
@@ -156,6 +157,31 @@ fn generate_secret_volume(id: &str) -> (Volume, VolumeMount) {
     )
 }
 
+pub async fn send_secret(client: Client, id: &str) -> Result<()> {
+    let secret_api: Api<Secret> = Api::default_namespaced(client);
+    let auth_secret = secret_api.get(TRUSTEE_AUTH_SECRET).await?;
```
alicefr, line 144: Why cannot we have this secreted mounted in the pod? This shouldn't be affected by the issue of dnymically adding and removing the screts

operator/src/trustee.rs
```diff
@@ -176,60 +158,26 @@ pub async fn send_secret(client: Client, id: &str) -> Result<()> {
...
-            info!("Secret {id} was to be dropped, but volume mount had already been removed");
-        }
-    }
-
-    deployments
-        .replace(TRUSTEE_DEPLOYMENT, &Default::default(), &deployment)
-        .await?;
+pub async fn delete_secret(client: Client, id: &str) -> Result<()> {
+    let secret_api: Api<Secret> = Api::default_namespaced(client);
+    let auth_secret = secret_api.get(TRUSTEE_AUTH_SECRET).await?;
```
alicefr: Same comment as above why not simply mounted the secret?

operator/src/trustee.rs
```diff
@@ -181,11 +179,89 @@ pub async fn delete_secret(client: Client, id: &str) -> Result<()> {
...
+    let auth_data = auth_secret.data.context("Auth secret has no data")?;
+    let auth_key_bytes = auth_data
+        .get("private.key")
+        .context("Auth secret missing private.key")?;
+    let auth_key = String::from_utf8(auth_key_bytes.0.clone())
+        .context("Auth private key is not valid UTF-8")?;
+
+    let url = format!("http://{TRUSTEE_SERVICE}:{TRUSTEE_PORT}");
```
alicefr: This looks like a good code to be refactored and put in a separate function since it repeates 3 times at least

alicefr:
I would love to see some integration tests here that creates multiple machine and try to restart the trustee pod and see if the secrets are correctly present also after the restart. Also a test where an secret was present before the restart of the trustee pod and that after a secret is dynamicall added and then deleted.

operator/src/main.rs
```diff
@@ -195,11 +197,24 @@ async fn install_trustee_configuration(
...
-    match trustee::generate_kbs_deployment(client, owner_reference, &trustee_image).await {
+    match trustee::generate_kbs_deployment(client.clone(), owner_reference, &trustee_image).await {
         Ok(_) => info!("Generate the KBS deployment"),
         Err(e) => error!("Failed to create the KBS deployment: {e}"),
     }
 
+    // Wait for the KBS pod to be ready before syncing secrets
+    let deployments: Api<Deployment> =
+        Api::default_namespaced(client.clone());
+    match trustee::wait_for_deployment_ready(&deployments).await {
```
alicefr: Looks more we need a reconcile loop for the deployment and that you sync the secret when its status is running. It doesn't seems the right place here to poll for a resource. Kubernetes is event driver and watch the resources

Containerfile
iroykaufman: The `openssl-sys` fails when this dependency is missing. This is because I bumped all the dependency versions, the upstream uses `openssl-sys 0.9.112`, and here it is `openssl-sys 0.9.113`. This is the error message:
```
cargo:warning=configuring OpenSSL build: 'perl' reported failure with exit status: 2
  cargo:warning=openssl-src: failed to build OpenSSL from source

  --- stderr
  Can't locate FindBin.pm in @INC (you may need to install the FindBin module) (@INC entries checked: /usr/local/lib64/perl5/5.42 /usr/local/share/perl5/5.42 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5) at ./Configure line 15.
  BEGIN failed--compilation aborted at ./Configure line 15.
```

operator/src/trustee.rs
iroykaufman: We can, but as I understand it from [issue#1260](https://github.com/confidential-containers/trustee/issues/1260), Trustee is moving from using the file system method for storing the secret and is suggesting the use of `kvstorage`, so using the API is a better approach that can be easier to support.

alicefr: Isn't this the certificate for accessing the kbs API? If so, then it isn't a secret to be added into KBS via the API

Jakob-Naucke: Hmm, the secret that's to be added to the KBS is below at l. 160 and that does seem right to me?

alicefr: Yes, what I mean it's that the private certificate to authenticate to the KBS can be a regular mounted secret. It doesn't need to be fetched by the secret API and parsed

Jakob-Naucke: You mean the operator should generate the secret, mount it to the operator pod (triggers a restart but sure, should be stable) and reconcile from there?

alicefr: why does this trigger a restart? This is the secret for the KBS so it should be unique and persistent. If the secret is rotated, then the update should be automatically done by kubernetes

operator/src/trustee.rs
```diff
@@ -346,6 +328,29 @@ pub async fn generate_secret(
...
+    let secret = Secret {
+        metadata: ObjectMeta {
+            name: Some(TRUSTEE_AUTH_SECRET.to_string()),
+            owner_references: Some(vec![owner_reference]),
+            ..Default::default()
+        },
+        data: Some(data),
+        ..Default::default()
+    };
+    create_or_info_if_exists!(client, Secret, secret);
```
Jakob-Naucke, line 352: @alicefr do I understand you correctly that you believe this secret should be created externally (e.g. `make` something) and mounted to the operator, instead of being created here?

alicefr: no, I don't want to be created externally but the secret can be created at start time and be mounted in the operator deployment. 

However, I don't quite like the idea that the operator update its self but I don't think it the authentication secret needs to be handled via the secret API

operator/src/trustee.rs
iroykaufman: The auth key for the KBS API is created by the operator and mount to the trustee pod. The creation of the key happened [here](https://github.com/trusted-execution-clusters/operator/pull/242/changes#diff-6916676a025aa3cf9969dd3c2f318a3603f8159611320d3759f9de0ae4cf6633R184) and it is mounted to the deployment [here](https://github.com/trusted-execution-clusters/operator/pull/242/changes/1d2bffc97f9704c4f6beeadd2cf5315fb419352d#diff-d6b9008c1b20f16862a6697389de650b674b54756b2a71bdc2fdf96cc1a56da4R503)

operator/src/trustee.rs
alicefr: For now, we decided to fetch the secret from the k8s api server to avoid any restart of the operator as it is already implemented by this PR

operator/src/trustee.rs
alicefr: See https://github.com/trusted-execution-clusters/operator/pull/242#discussion_r3109305282

operator/src/main.rs
iroykaufman: Done

operator/src/kbs-config.toml
```diff
@@ -4,6 +4,7 @@ insecure_http = true
 
 [admin]
 insecure_api = true
+auth_public_key="/key/public.pub"
```
alicefr, line 7: nit: space between the 2 string and the =

operator/Cargo.toml
```diff
@@ -30,6 +30,7 @@ serde.workspace = true
 serde_json.workspace = true
 thiserror = "2.0.18"
 tokio.workspace = true
+kbs-client = {git = "https://github.com/iroykaufman/trustee/", branch = "ak-registration"}
```
alicefr, line 33: Any reason for using your fork?

iroykaufman: Yes, currently registering AK using the KBS API is not supported. I created this [PR#1306](https://github.com/confidential-containers/trustee/pull/1306) to include support.

operator/src/trustee.rs
```diff
@@ -181,11 +179,89 @@ pub async fn delete_secret(client: Client, id: &str) -> Result<()> {
...
+        .iter()
+        .filter(|secret| {
+            secret.metadata.deletion_timestamp.is_none()
+                && secret
+                    .metadata
+                    .owner_references
+                    .as_ref()
+                    .map(|owners| owners.iter().any(|owner| owner.kind == "Machine"))
+                    .unwrap_or(false)
+        })
```
alicefr, line 244: Instead of filtering by owner, we can introduce a label and mark the secret owned by a machine. In this way, the filtering is doing by kubernetes api

operator/src/trustee.rs
```diff
@@ -257,6 +218,47 @@ pub async fn sync_all_machine_secrets(client: Client) -> Result<()> {
...
+    client: Arc<Client>,
+) -> Result<Action, ControllerError> {
+    if let Some(status) = &deployment.status {
+        let desired = deployment
+            .spec
+            .as_ref()
+            .and_then(|s| s.replicas)
+            .unwrap_or(1);
+        let ready = status.ready_replicas.unwrap_or(0);
+        if ready >= desired && desired > 0 {
```
alicefr, line 232: why do we need the desired? Cannot we simply evaluate `ready > 1` ?

tests/trusted_execution_cluster.rs
```diff
@@ -434,3 +436,383 @@ async fn test_attestation_key_lifecycle() -> anyhow::Result<()> {
...
+        "spec": {
+            "template": {
+                "metadata": {
+                    "annotations": {
+                        "kubectl.kubernetes.io/restartedAt": now
+                    }
+                }
+            }
+        }
+    });
```
alicefr, line 555: cannot we simply delete the kbs pod?

alicefr:
can we also test the deletion of a secret  in one of the integration tests

PR #248, opened 2026-05-04
operator: KBS API for LUKS key registration

### LINKED ISSUE #229: Dynamically add and remove secret from the trustee pod
Currently, the flow for creating a LUKS key is:
1. When the operator detects new machine creation, it creates a new LUKS key.
2. The operator creates a Kubernetes secret to hold the LUKS key.
3. Patch the trustee deployment to add the secret as a confidential resource.

Step 3 will trigger the trustee deployment to restart with the new secret mounted. We want to avoid this. Other objects (policies, reference-values, etc) can be handled similarly.

I start implementing two different approaches to solving this issue:

1. Create a sidecar inside the trustee deployment that listens for new machine creation and manages the LUKS key as a Kubernetes secret. A draft implementation can be found [here](https://github.com/trusted-execution-clusters/operator/compare/main...iroykaufman:operator:luks-sidecar). Based on my tests, this implementation is functional.
2. Use the kbs API (`kbs-client`) from trustee to update the LUKS key and other trustee configuration. While the first two steps of the flow remain the same, the third step is replaced by a direct API call to the KBS. A draft implementation is available [here](https://github.com/trusted-execution-clusters/operator/compare/main...iroykaufman:operator:kbs-api). This version is currently incomplete and has not been fully tested.

I would like to get your feedback on what is a better approach

iroykaufman:
Currently, every time the LUKS key is updated, the operator patches the trustee deployment, which causes a restart of the pod. This PR introduces a way to avoid this by setting the LUKS key using the KBS API. 

## Core implementation points:
* Right before the trustee first deploys the operator, create an auth key for the KBS API and save it as a kube secret.
* Replacing the patch mechanism with API calls
* Add a reconcile loop that watches for changes in the trustee deployment and sync LUKS key that was lost.

## Tests:
* test_luks_key_sync - this test checks that the luks key is first sent to tustee and validates that after trustee restarts, the keys are sent again. Also, at the end, I delete one of the machines and check that the secret has been deleted.

## Summary by Sourcery

Integrate KBS API–based LUKS key management for trustee, replacing deployment patching with authenticated API calls and adding automatic resync on trustee restarts.

New Features:
- Introduce authenticated KBS API integration to store and manage per-machine LUKS keys instead of mounting secrets into the trustee deployment.
- Add a controller that watches the trustee deployment and re-synchronizes all machine LUKS keys to KBS when the trustee becomes ready.
- Generate and manage an Ed25519 keypair as a Kubernetes secret for authenticating KBS API access from the operator.

Enhancements:
- Label and configure the trustee (KBS) service and deployment consistently via a shared label selector constant and updated volume templates.
- Switch KBS resource storage to the kvstorage plugin and configure KBS admin to use the public auth key for API access.

Build:
- Pin compute-pcrs-lib to a specific Git revision for reproducible builds and add kbs-client as a new operator dependency, along with installing perl in the container image for the build tooling.

Tests:
- Add an end-to-end test that verifies initial LUKS key upload, re-sync after trustee restart, and LUKS key deletion on machine removal, plus unit tests for Ed25519 key generation.


operator/src/trustee.rs
```diff
@@ -120,74 +142,113 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
+}
 
-    deployments
-        .replace(TRUSTEE_DEPLOYMENT, &Default::default(), &deployment)
-        .await?;
-    Ok(())
+pub async fn launch_trustee_sync_controller(client: Client) {
+    let deployments: Api<Deployment> = Api::default_namespaced(client.clone());
+    let watcher_config = watcher::Config {
+        label_selector: Some("app=kbs".to_string()),
```
Jakob-Naucke: nit: maybe a constant also used in `generate_kbs_{service,deployment}` is better

operator/src/trustee.rs
```diff
@@ -18,15 +18,20 @@ use k8s_openapi::apimachinery::pkg::{
...
 };
-use log::info;
-use operator::{RvContextData, create_or_info_if_exists};
+use log::{info, warn};
+use operator::{ControllerError, RvContextData, controller_error_policy, controller_info, create_or_info_if_exists};
+use std::sync::Arc;
 use serde::{Serialize, Serializer};
 use serde_json::{Value::String as JsonString, json};
 use std::collections::BTreeMap;
+use kbs_client;
```
Jakob-Naucke: Is this going to be https://github.com/confidential-containers/trustee/tree/main/tools/kbs-client?

Jakob-Naucke (commented):
Thank you for creating this, and thank you for already creating a test. Make sure that linting & build/unit tests pass.

operator/src/trustee.rs
iroykaufman: Right

iroykaufman:
The CI is failing because we need to have some Perl dependency. I solved this by adding this `RUN dnf install -y perl-FindBin perl-core` to the containerfile.  @Jakob-Naucke, any idea what the best approach is to solve it for the CI?

Jakob-Naucke:
@iroykaufman these tests run in (Ubuntu) GHA containers that aren't affected by the Containerfile (this would get picked up in integration tests once I manage to fix them). You'll need to add them to the steps in `.github/rust.yml`:
```yaml
steps:
  - name: "Install OpenSSL dependencies"
    run: apt-get install -y …
```

Jakob-Naucke:
@iroykaufman whoops, you can set your own build container, and we do, and it's a Fedora container, so it should be `dnf` as usual. You did set it in the good place though afaict.

Jakob-Naucke:
@iroykaufman thanks for the updates. integration tests are failing because these packages are missing there too. in my intuition, those shouldn't need to be built on the host at all so let me look into that first.

e: before the next push you can also still look into the lint failures

iroykaufman:
> @iroykaufman thanks for the updates. integration tests are failing because these packages are missing there too. in my intuition, those shouldn't need to be built on the host at all so let me look into that first.

Thanks, let me know and then I'll push with the lint fix

iroykaufman:
@Jakob-Naucke  maybe update the container file in [buildroot](https://github.com/trusted-execution-clusters/buildroot) with the openssl dependency is better then adding them in the github workflow files. WDYT?

Jakob-Naucke:
> @Jakob-Naucke maybe update the container file in [buildroot](https://github.com/trusted-execution-clusters/buildroot) with the openssl dependency is better then adding them in the github workflow files. WDYT?

Yes. I got confused by first thinking we were on GHA's Ubuntu containers, but the buildroot is the better place.

Jakob-Naucke:
and the CI host does need these to compile the integration tests, which is slightly not how I thought cargo dependencies worked, but alas, these packages are now installed

iroykaufman:
I deleted the changes for the github workflow and opened [PR#9](https://github.com/trusted-execution-clusters/buildroot/pull/9). When this is merged, the CI should work

alicefr:
@sourcery-ai review

operator/src/kbs-config.toml
```diff
@@ -2,7 +2,12 @@
 sockets = ["0.0.0.0:8080"]
 
 [admin]
-type = "DenyAll"
+type = "Simple"
```
Jakob-Naucke: Sorry about this one, didn't see it coming but you fixed it :)

Jakob-Naucke (commented):
- If changes to the base branch made a rebase necessary, the usual practice is to fix and keep the existing commits, so I think you can squash in this instance.
- Pushing new buildroots got auto-disabled due to inactivity (the more you know). I re-enabled it so FindBin etc. are hopefully there on next attempt.

iroykaufman:
>If changes to the base branch made a rebase necessary, the usual practice is to fix and keep the existing commits, so I think you can squash in this instance.

This is still WIP so I'll squash it into one commit after everything is fixed.

alicefr:
@iroykaufman can you please split the git history into multiple commits to make the review easier

iroykaufman:
I see that we disallow the use of `rustls`, and I understand that we prefer `openssl`. `kbs-client` has a dependency that uses `rustls`  so to use the kbs API we need to allow the use of `rustls`.  @Jakob-Naucke @alicefr WDYT?

tests/trusted_execution_cluster.rs
```diff
@@ -434,3 +439,203 @@ async fn test_attestation_key_lifecycle() -> anyhow::Result<()> {
...
+    // Delete machine1 and verify its secret is removed from both K8s and KBS
+    machines
+        .delete(&machine1_name, &Default::default())
+        .await?;
+    test_ctx.info(format!("Deleted Machine 1: {machine1_name}"));
+
+    let poller = Poller::new()
+        .with_timeout(Duration::from_secs(60))
+        .with_interval(Duration::from_secs(2))
+        .with_error_message("Machine1 secret not deleted from KBS".to_string());
```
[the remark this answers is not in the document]
alicefr: Yes, please don't check the trustee log, it is a bad practice. But rather, try to evaluate if the objects exist/have been deleted

Jakob-Naucke:
> I see that we disallow the use of `rustls`, and I understand that we prefer `openssl`. `kbs-client` has a dependency that uses `rustls` so to use the kbs API we need to allow the use of `rustls`.

We have this lint because especially in downstream we want to use system SSL for FIPS & PQC, and that would extend to us using `kbs-client`. Sorry for the back and forth, but could you contribute a `native-tls` feature to CoCo upstream? I think some folks downstreaming from them would mutually benefit from such a feature.

iroykaufman:
> > I see that we disallow the use of `rustls`, and I understand that we prefer `openssl`. `kbs-client` has a dependency that uses `rustls` so to use the kbs API we need to allow the use of `rustls`.
> 
> We have this lint because especially in downstream we want to use system SSL for FIPS & PQC, and that would extend to us using `kbs-client`. Sorry for the back and forth, but could you contribute a `native-tls` feature to CoCo upstream? I think some folks downstreaming from them would mutually benefit from such a feature.

Yes, I'll try to do that

iroykaufman:
> @iroykaufman can you please split the git history into multiple commits to make the review easier

I split it into 4 commits. Hopefully, now it will be easier to review

operator/src/trustee.rs
```diff
@@ -137,74 +148,145 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
-        },
-        VolumeMount {
-            name: id.to_string(),
-            mount_path: format!("{TRUSTEE_SECRETS_PATH}/{id}"),
-            ..Default::default()
-        },
-    )
+    Ok((
+        format!(
+            "http://{}",
```
alicefr, line 172: why not https here as well?

operator/src/trustee.rs
```diff
@@ -137,74 +148,145 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
 }
 
-pub async fn unmount_secret(client: Client, id: &str) -> Result<()> {
-    let result = do_mount_secret(client, id, false).await;
-    info!("Unmounted secret {id} from {TRUSTEE_DEPLOYMENT}");
-    result
+pub async fn delete_secret(client: Client, id: &str) -> Result<()> {
+    let auth_key = get_auth_key(&client).await?;
+    let (url, certs) = get_kbs_connection(&client).await?;
+    let path = format!("default/{id}/root");
```
alicefr: can we create a more "official" functions that defines this path. I know it was already like this but since you are touching the code. Please add it in a separate commit

operator/src/trustee.rs
```diff
@@ -137,74 +148,145 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
+    info!("Secret {id} deleted successfully");
+    Ok(())
 }
 
-pub async fn do_mount_secret(client: Client, id: &str, add: bool) -> Result<()> {
-    let deployments: Api<Deployment> = Api::default_namespaced(client);
-    let mut deployment = deployments.get(TRUSTEE_DEPLOYMENT).await?;
+pub async fn sync_all_machine_luks_key(client: Client) -> Result<()> {
+    let secret_api: Api<Secret> = Api::default_namespaced(client.clone());
+    let secret_list = secret_api.list(&Default::default()).await?;
```
alicefr: can we add a label to the secret that is owned by a machine. In this way, we can filter the secrets when we query them from the API server. It saves some memory since we don't need to store all the secret on the operator side

operator/src/trustee.rs
iroykaufman: It's http only when no cert exists.

operator/src/trustee.rs
iroykaufman: I just want to make sure I understand correctly. Create a function that takes an id and returns `format!("default/{id}/root")` ?

alicefr: yes, so it can be reused eventually

Jakob-Naucke:
note: this also unblocks bumping Trustee, where a lot of config is better set via API since https://github.com/confidential-containers/trustee/pull/1190

Jakob-Naucke:
ooh the trustee bump might not work as is. I played with it yesterday because I needed a new image anyhow, but for this to work, I think you'll need a new guest image, new configs, and setting policies with the client. I think it's out of scope for this PR.

iroykaufman:
> ooh the trustee bump might not work as is. I played with it yesterday because I needed a new image anyhow, but for this to work, I think you'll need a new guest image, new configs, and setting policies with the client. I think it's out of scope for this PR.

I updated the `kbs-config` file to align with the latest version since I needed to modify it anyway, and this is working for me. Will the prow CI automatically use the v0.20.0 image since it's set as the default in the Makefile?

Jakob-Naucke:
> I updated the `kbs-config` file to align with the latest version since I needed to modify it anyway, and this is working for me. Will the prow CI automatically use the v0.20.0 image since it's set as the default in the Makefile?

Yes, but by "new image" I meant a new `TEST_IMAGE` with the appropriate guest components (although maybe those could work -- I updated gc first and that needed new Trustee, but maybe it's still forwards compatible).

Good for updating the kbs-config, but looking at the PR I linked, e.g. `policy_engine` was removed right?

e: but since you pushed we can see what happens with CI

iroykaufman:
>Yes, but by "new image" I meant a new TEST_IMAGE with the appropriate guest components (although maybe those could work -- I updated gc first and that needed new Trustee, but maybe it's still forwards compatible).

Right, I missed that the attestation test will likely fail. What do you think about I'll open a new PR to bump trustee to v0.20.0 instead? I've already added support for the new changes for that version here anyway.

iroykaufman:
>Good for updating the kbs-config, but looking at the PR I linked, e.g. policy_engine was removed right?

Right, its look like the `opa`  was removed but I'm still now sure what should go instead.

iroykaufman:
The CI pass that's good, so the kbs-client is compatible with this version of trustee, at least for this use case.

operator/src/trustee.rs
iroykaufman: I changed the implementation to go over the machines and then get the secret by the machine UID. Let me know if this is fine too

operator/src/trustee.rs
```diff
@@ -346,6 +363,35 @@ pub async fn generate_secret(
...
 }
 
+pub async fn generate_trustee_auth_keys_secret(
+    client: Client,
+    owner_reference: OwnerReference,
+) -> Result<()> {
+    let key_pair = generate_ed25519_key_pair()?;
+    let data = BTreeMap::from([
+        (
+            "private.key".to_string(),
```
Jakob-Naucke: Please `const` this and `public.pub`

operator/Cargo.toml
```diff
@@ -31,6 +31,11 @@ serde_json.workspace = true
 thiserror = "2.0.18"
 tokio.workspace = true
 toml = "1.1.2"
+kbs-client = {git = "https://github.com/iroykaufman/trustee.git", branch = "kbs-client-native-tls", default-features = false, features = ["native-tls"] }
+jsonwebtoken = { version = "10.4.0", default-features = false, features = [
+    "aws_lc_rs",
```
Jakob-Naucke: @uril don't we also need to add a FIPS feature to jsonwebtoken to use https://github.com/aws/aws-lc-rs/blob/231ca7db25d86ade03b8edb19efffee5bfb59fbe/aws-lc-rs/Cargo.toml#L48?

operator/src/kbs-config.toml
```diff
@@ -30,8 +37,7 @@ policy_engine = "opa"
 
 [[plugins]]
 name = "resource"
-type = "LocalFs"
-dir_path = "/opt/trustee/kbs-repository"
+storage_backend_type = "kvstorage"
```
Jakob-Naucke, line 30: Should the Trustee bump in the same commit?

operator/src/trustee.rs
```diff
@@ -137,74 +149,140 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
+    let secret_api: Api<Secret> = Api::default_namespaced(client.clone());
+    let auth_secret = secret_api.get(TRUSTEE_AUTH_SECRET).await?;
+    let auth_data = auth_secret.data.context("Auth secret has no data")?;
+    let auth_key_bytes = auth_data
+        .get("private.key")
+        .context("Auth secret missing private.key")?;
+
+    let claims = json!({
+        "role": "admin",
+        "exp": 2147483647 // Set to Jan 19, 2038
```
Jakob-Naucke: nit: use `i32::MAX`

operator/src/trustee.rs
```diff
@@ -137,74 +149,140 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
+    client: Arc<Client>,
+) -> Result<Action, ControllerError> {
+    if let Some(status) = &deployment.status {
+        let desired = deployment
+            .spec
+            .as_ref()
+            .and_then(|s| s.replicas)
+            .unwrap_or(1);
+        let ready = status.ready_replicas.unwrap_or(0);
+        if ready >= desired && desired > 0 {
```
Jakob-Naucke: Could race, better check if the deployment is `Available` instead

tests/trusted_execution_cluster.rs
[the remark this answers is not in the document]
Jakob-Naucke: ig this is still open

Cargo.toml
```diff
@@ -18,7 +18,7 @@ axum-server = { version = "0.8.0", features = ["tls-openssl"] }
 chrono = "0.4.44"
 clap = { version = "4.6.1", features = ["derive"] }
 clevis-pin-trustee-lib = { git = "https://github.com/latchset/clevis-pin-trustee" }
-compute-pcrs-lib = { git = "https://github.com/trusted-execution-clusters/compute-pcrs" }
+compute-pcrs-lib = { git = "https://github.com/trusted-execution-clusters/compute-pcrs", rev = "1e7b9f74206e436d1426c335e30b2f1a6bd1681e"}
```
Jakob-Naucke: This change seems unnecessary as we already use that version. The Cargo.lock updates from this commit should be in the same commit as the Cargo.toml that changed them.

operator/src/trustee.rs
```diff
@@ -200,6 +200,10 @@ async fn get_kbs_connection(client: &Client) -> Result<(String, Vec<String>)> {
     ))
 }
 
+pub fn secret_path(id: &str) -> String {
```
Jakob-Naucke, line 326: can squash this commit

operator/src/kbs-config.toml
```diff
@@ -30,8 +37,7 @@ policy_engine = "opa"
...
 
 [policy_engine]
 policy_path = "/opt/trustee/policy.rego"
```
Jakob-Naucke: > policy_engine was removed right?

Yes. E.g. removing the `allow if` block from `operator/src/resource.rego` still gives a pass for me.
Do you need the updates from Trustee v0.20.0 here? If yes, you'd either need to first go back to v0.17.0 compatibility or set the policies using kbs-client too, which is now the only supported way AFAICT. If no, we should bump Trustee in a later PR.

Also not for this PR but if we were being pedantic we'd ensure that the policies get picked up…

operator/src/trustee.rs
iroykaufman: Let's check with @alicefr first, since she suggested putting it in a separate commit ([here](https://github.com/trusted-execution-clusters/operator/pull/248#discussion_r3309509862)).  For me, it's fine either way.

Jakob-Naucke: ah right, but then I think that commit should come first

operator/Cargo.toml
uril: Yes, that would be good.

I think that it would have been even better if the crate enabled openssl

operator/src/trustee.rs
iroykaufman: OK make sense.

operator/Cargo.toml
Jakob-Naucke: @iroykaufman could you look into contributing that? sorry for all the back and forth

operator/src/kbs-config.toml
iroykaufman: I have to use at least Trustee v0.19.0 for `kbs_client::delete_resource` to be supported. In order to bump to Trustee v0.20.0 correctly, [PR#273](https://github.com/trusted-execution-clusters/operator/pull/273) is also needed because now the AK is checked against the reference value. I think it would be best to bump to Trustee v0.19.0 in this PR and later bump into v0.20.0 WDYT?

operator/Cargo.toml
iroykaufman: @uril @Jakob-Naucke I opened that [PR](https://github.com/Keats/jsonwebtoken/pull/518) on jsonwebtoken

Jakob-Naucke: @iroykaufman sorry, I missed it too that it already existed :/ but it seems usable

iroykaufman: It's OK, this task looked very fit for AI so I let it do most of the job anyway :sweat_smile:

operator/src/kbs-config.toml
Jakob-Naucke: sounds good, sorry for delayed response

iroykaufman: How can I build the Trustee v0.19.0 image like the one on quay.io/trusted-execution-clusters/key-broker-service ?

Jakob-Naucke: I did `podman build --build-arg COMMIT=v0.20.0 -t quay.io/trusted-execution-clusters/key-broker-service:v0.20.0 -f trustee.container` from `investigations/containerfiles`, then you should be able to use 19 instead

Jakob-Naucke: that containerfile points to our fork, not sure if that tag is there. if not, you can use upstream instead or replicate that tag there

Jakob-Naucke (commented):
To not break bisects, the Trustee bump should be in the same commit as the API interaction right? It's good that the tests remain separate like you put it though.

iroykaufman:
> To not break bisects, the Trustee bump should be in the same commit as the API interaction right? It's good that the tests remain separate like you put it though.

The API should also work with v0.17.0, but we are using a newer version of kbs-client this can be a problem, so I'll put them in the same commit.  

I need to put the three last commits(without the tests) together hupfuly is will be clear to read. If you have any other ideas to make it better, let me know

operator/src/trustee.rs
```diff
@@ -42,6 +42,10 @@ pub(crate) const TRUSTEE_DATA_MAP: &str = "trustee-data";
 const ATT_POLICY_MAP: &str = "attestation-policy";
 const TRUSTED_AK_KEYS_VOLUME: &str = "trusted-ak-keys";
 const TRUSTED_AK_KEYS_DIR: &str = "/etc/tpm/trusted_ak_keys";
+const TRUSTEE_AUTH_SECRET: &str = "trustee-auth";
+const TRUSTEE_AUTH_KEY_DIR: &str = "/key";
```
uril: nit: perhaps better /opt/trustee/keys or /etc/trustee/keys

operator/src/kbs-config.toml
```diff
@@ -1,37 +1,36 @@
 [http_server]
-sockets = ["0.0.0.0:8080"]
+  sockets = ["0.0.0.0:8080"]
```
uril: There are no spaces added in a toml file.

operator/src/main.rs
```diff
@@ -160,9 +160,9 @@ async fn install_trustee_configuration(
         Err(e) => error!("Failed to create the KBS configuration configmap: {e}"),
     }
 
-    match trustee::generate_attestation_policy(client.clone(), owner_reference.clone()).await {
-        Ok(_) => info!("Generate configmap for the attestation policy",),
-        Err(e) => error!("Failed to create the attestation policy configmap: {e}"),
+    match reference_values::create_pcrs_config_map(client.clone()).await {
```
uril: Is there a need to keep the call to reference_values::create_pcrs_config_map in main() (line -291 or +297)?

iroykaufman: There is no need for this call in main. I removed it

uril (commented):
Consider to rephrased for this commit:
Subject: Fix race of ApprovedImage adoption with TEC object deletion

The race scenario:
- Assume the operator is installed and TEC object exists
- Apply an approved-image yaml file (create a new CR instance)
- Delete the TEC object (immediately)
- The new approved-image is not deleted with the TEC object

This happens because <what you wrote>

Jakob-Naucke:
image push failed? I hope #287 stabilizes things a bit

/retest

operator/src/reference_values.rs
```diff
@@ -247,6 +232,12 @@ async fn image_reconcile(
         .await
         .map_err(|e| -> ControllerError { e.into() })?;
 
+    if let Some(ref cluster) = cluster {
+        adopt_approved_image(kube_client.clone(), &name, cluster)
```
Jakob-Naucke, line 262: Good catch, but should we still check if it's owned?

operator/src/main.rs
```diff
@@ -15,7 +15,7 @@ use kube::runtime::controller::{Action, Controller};
 use kube::runtime::reflector::{self, Store};
 use kube::runtime::watcher;
 use kube::{Api, Client};
-use log::{info, warn, error};
+use log::{error, info, warn};
```
Jakob-Naucke: Please run lints after every commit

tests/trusted_execution_cluster.rs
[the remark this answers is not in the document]
Jakob-Naucke: ig this is still open

tests/trusted_execution_cluster.rs
```diff
@@ -369,6 +372,437 @@ async fn test_nonexistent_approved_image() -> anyhow::Result<()> {
     let ctx = "waiting for ApprovedImage coreos1 to be PodPending";
     timeout(scaled_duration(30), done).await.context(ctx)??;
 
+     test_ctx.cleanup().await?;
+    Ok(())
+}
+}
+
+named_test! {
+async fn test_luks_key_sync() -> anyhow::Result<()> {
```
Jakob-Naucke, line 381: Please include a more descriptive commit message subject than "Add tests", e.g. `Add test_{luks,attestation}_key_sync`

Jakob-Naucke (commented):
Looks very promising. A few small comments.

tests/trusted_execution_cluster.rs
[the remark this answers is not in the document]
iroykaufman: I add a check that the machine Secret is deleted from Kubernetes later in the test. I prefer to keep both checks so I verify that the secret was deleted from trustee and Kubernetes.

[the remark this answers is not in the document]
Jakob-Naucke: ah cool, I see it now

tests/trusted_execution_cluster.rs
```diff
@@ -369,6 +372,437 @@ async fn test_nonexistent_approved_image() -> anyhow::Result<()> {
...
+                    Err(kube::Error::Api(ae)) if ae.code == 404 => Ok(()),
+                    Err(e) => Err(e.into()),
+                }
+        }
+    })
+    .await?;
+    test_ctx.info("Machine1 secret deleted from KBS");
+
+    // Verify the K8s Secret for machine1 is also deleted
+    wait_for_resource_deleted(&secrets_api, &machine1_uuid, 60).await?;
```
Jakob-Naucke: You don't need the poller when you also check with `wait_for_resource_deleted`, right?

Jakob-Naucke (commented):
In the tests, please also used scaled_timeout like in #276

operator/src/reference_values.rs
iroykaufman: I'll add this check, this can save some unnecessary patches.

tests/trusted_execution_cluster.rs
iroykaufman: Right, my bad I missed that when I applied sourcery suggestion.

Jakob-Naucke (dismissed):
LGTM, cc @alicefr

tests/trusted_execution_cluster.rs
```diff
@@ -369,6 +372,424 @@ async fn test_nonexistent_approved_image() -> anyhow::Result<()> {
...
+    // Apply the patch
+    deployments
+        .patch(
+            TRUSTEE_DEPLOYMENT,
+            &PatchParams::default(),
+            &Patch::Strategic(patch),
+        )
+        .await?;
+
+    test_ctx.wait_for_deployment_ready(&deployments, TRUSTEE_DEPLOYMENT, 120).await?;
```
alicefr, line 493: Instead of polling here, can we  replace it with await_condition?
Something like:
```rust
  let is_available = |obj: Option<&Deployment>| -> bool {
      obj.and_then(|d| d.status.as_ref())
          .map(|r| r >= 1)
          .unwrap_or(false)
  };
  let ready = await_condition(deployments.clone(), TRUSTEE_DEPLOYMENT, is_available);
  timeout(Duration::from_secs(120), ready).await.context("waiting for trustee to be ready")??;
```

alicefr: @Jakob-Naucke probably we could replace several polling operation with that. WDYT?

tests/trusted_execution_cluster.rs
```diff
@@ -369,6 +372,424 @@ async fn test_nonexistent_approved_image() -> anyhow::Result<()> {
...
+    test_ctx.info(format!("Created AttestationKey 2: {ak2_name}"));
+
+    // Wait for both AKs to be approved and have secrets created
+    let secrets_api: Api<Secret> = Api::namespaced(client.clone(), namespace);
+    let poller = Poller::new()
+        .with_timeout(Duration::from_secs(60))
+        .with_interval(Duration::from_millis(500))
+        .with_error_message("AttestationKeys not approved with secrets".to_string());
+
+    poller
```
alicefr: This probably can be refactor with the await_condition

alicefr: and the same above using the polling

tests/trusted_execution_cluster.rs
Jakob-Naucke: ah yes, I missed these

operator/src/trustee.rs
```diff
@@ -122,214 +188,231 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
-        .map(|vm| {
-            vm.iter()
-                .filter(|mount| mount.name != TRUSTED_AK_KEYS_VOLUME)
-                .cloned()
-                .collect()
+        .filter_map(|secret| {
+            secret
+                .data
+                .as_ref()
+                .and_then(|d| String::from_utf8(d.get("public_key").unwrap().0.clone()).ok())
```
alicefr: what if the secret doesn't have the field public_key, this will panic:
```suggestion
  .filter_map(|secret| {
      secret
          .data
          .as_ref()
          .and_then(|d| d.get("public_key"))
          .and_then(|pk| String::from_utf8(pk.0.clone()).ok())
  })
```

operator/src/main.rs
```diff
@@ -167,10 +165,21 @@ async fn install_trustee_configuration(
...
     info!("Generated configmap for the KBS configuration");
 
-    trustee::generate_attestation_policy(client.clone(), owner_reference.clone())
-        .await
-        .context("Failed to create the attestation policy configmap")?;
-    info!("Generated configmap for the attestation policy");
+    match reference_values::create_pcrs_config_map(client.clone()).await {
+        Ok(_) => info!("Created bare configmap for PCRs"),
+        Err(e) => error!("Failed to create the PCRs configmap: {e}"),
+    }
```
alicefr: This changes the previous behavior where we were returning an error

operator/src/main.rs
```diff
@@ -167,10 +165,21 @@ async fn install_trustee_configuration(
...
+    match trustee::generate_trustee_auth_keys_secret(client.clone(), owner_reference.clone()).await
+    {
+        Ok(_) => info!("Generate auth keys for the KBS API",),
+        Err(e) => error!("Failed to create the auth keys: {e}"),
+    }
+
+    match trustee::generate_rv_data(client.clone(), owner_reference.clone()).await {
+        Ok(_) => info!("Created configmap for reference values"),
+        Err(e) => error!("Failed to create the reference values configmap: {e}"),
+    }
```
alicefr: Should here return an error instead of just printing it?

operator/src/trustee.rs
```diff
@@ -122,214 +188,231 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
+        let c = Arc::unwrap_or_clone(client.clone());
+        if let Err(e) = sync_resource_policy(c.clone()).await {
+            warn!("Failed to sync resource policy to KBS: {e}");
+        }
+        if let Err(e) = sync_attestation_policy(c.clone()).await {
+            warn!("Failed to sync attestation policy to KBS: {e}");
         }
+        if let Err(e) = sync_reference_values_from_configmap(&c).await {
+            warn!("Failed to sync reference values to KBS: {e}");
+        }
```
alicefr, line 276: why if those fails we only print a warning and not return an error?

operator/src/trustee.rs
```diff
@@ -122,214 +188,231 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
 
-pub async fn update_attestation_keys(ctx: &AkContextData) -> Result<()> {
-    let client = &ctx.client;
-    let ak_secrets: Vec<String> = ctx
-        .secret_store
-        .state()
-        .into_iter()
+pub async fn update_attestation_keys(client: Client) -> Result<()> {
+    let secrets: Api<Secret> = Api::default_namespaced(client.clone());
+    let secret_list = secrets.list(&Default::default()).await?;
```
alicefr, line 403: This doesn't use the secret store any longer. This was introduced by https://github.com/trusted-execution-clusters/operator/pull/267 to avoid direct APi calls to the API server

operator/src/reference_values.rs
```diff
@@ -563,36 +551,6 @@ mod tests {
         test_error_method!(clos, Method::PATCH);
     }
 
-    #[tokio::test]
-    async fn test_adopt_approved_images() {
```
alicefr, line 571: Why those tests don't apply any longer?

operator/src/trustee.rs
```diff
@@ -141,214 +188,231 @@ fn generate_luks_key() -> Result<Vec<u8>> {
     };
     serde_json::to_vec(&jwk).map_err(Into::into)
 }
+async fn get_auth_key_token(client: &Client) -> Result<String> {
+    let secret_api: Api<Secret> = Api::default_namespaced(client.clone());
+    let auth_secret = secret_api.get(TRUSTEE_AUTH_SECRET).await?;
```
alicefr, line 134: Ideally, this should use the secret store and not calling the API server directly

operator/src/trustee.rs
```diff
@@ -141,214 +188,231 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
+
+    let encoding_key = EncodingKey::from_ed_pem(auth_key_bytes.0.as_slice())?;
+
+    let token = encode(&Header::new(Algorithm::EdDSA), &claims, &encoding_key)?;
+    Ok(token)
+}
+async fn get_kbs_connection(client: &Client) -> Result<(String, Vec<String>)> {
+    let tec = trusted_cluster_operator_lib::get_trusted_execution_cluster(client.clone()).await?;
+    let secret_api: Api<Secret> = Api::default_namespaced(client.clone());
+
```
alicefr, line 153: Same here, please use the secret store

operator/src/main.rs
Jakob-Naucke: :+1: I know it wasn't out when this was raised but we changed this in #285

alicefr:
@iroykaufman can you please also split the commit https://github.com/trusted-execution-clusters/operator/pull/248/changes/774f99bba3467ab64ae03b0d75eb5a33f3534d2b into multiple It does too many things in one. I would at least split the bump of the kbs-client version and the configuration into a separate commit. Also I would separate the testing in its own commit. Then, it becomes easier to read

alicefr:
I would also personally move the https://github.com/trusted-execution-clusters/operator/pull/248/changes/512baf6bf1235adeb81a486143a624286e4053c4 into a separate PR. It really solves a bug and it isn't part of this feature

Jakob-Naucke:
> @iroykaufman can you please also split the commit [774f99b](https://github.com/trusted-execution-clusters/operator/commit/774f99bba3467ab64ae03b0d75eb5a33f3534d2b) into multiple It does too many things in one. I would at least split the bump of the kbs-client version and the configuration into a separate commit.

@iroykaufman was the problem that the kbs-client won't work as used here with 17? if it does work, I agree. if it doesn't, I prefer big commits over breaking bisects @alicefr

operator/src/main.rs
uril: Please fix similarly the call to generate_trustee_auth_keys_secret() above

operator/src/trustee.rs
```diff
@@ -111,6 +115,21 @@ pub async fn update_reference_values(client: Client) -> Result<()> {
     Ok(())
 }
 
+pub struct Ed25519KeyPair {
+    pub private_key_pem: Vec<u8>,
+    pub public_key_pem: Vec<u8>,
+}
+
+fn generate_ed25519_key_pair() -> Result<Ed25519KeyPair> {
+    let key = openssl::pkey::PKey::generate_ed25519()?;
```
uril, line 305: For now it's OK, but we should later prefer PQC keys (if available) over Ed25519

tests/trusted_execution_cluster.rs
uril: Don't we need to look for a condition with ( type==Available && status == "True" ) ?
Alternatively we can check for r.available_replicas >= 1

operator/Cargo.toml
```diff
@@ -34,6 +34,7 @@ toml = "1.1.2"
 kbs-client = {git = "https://github.com/confidential-containers/trustee.git", rev = "e65897a9ad4eb3ac69fa2ec75ed831200eb2acd7", default-features = false, features = ["native-tls"] }
 jsonwebtoken = { version = "10.4.0", default-features = false, features = ["use_pem"] }
 jsonwebtoken-openssl = "1.0.0"
+jiff = "0.2.29"
```
uril: nit: can we use chrono instead of jiff (for deletion_timestamp) -- it's already required.

operator/src/reference_values.rs
iroykaufman: Because we no longer need the function `adopt_approved_images`, we only use and test (above) `adopt_approved_image`

iroykaufman:
> > @iroykaufman can you please also split the commit [774f99b](https://github.com/trusted-execution-clusters/operator/commit/774f99bba3467ab64ae03b0d75eb5a33f3534d2b) into multiple It does too many things in one. I would at least split the bump of the kbs-client version and the configuration into a separate commit.
> 
> @iroykaufman was the problem that the kbs-client won't work as used here with 17? if it does work, I agree. if it doesn't, I prefer big commits over breaking bisects @alicefr

Correct. When bumping to trustee v0.20.0, it didn't work as is because the trustee storage backend changed. The idea was to use the API as an abstraction layer that could make it much easier to bump trustee to a later version or change the storage backend in the future.

iroykaufman:
> I would also personally move the [512baf6](https://github.com/trusted-execution-clusters/operator/commit/512baf6bf1235adeb81a486143a624286e4053c4) into a separate PR. It really solves a bug and it isn't part of this feature

I created this [PR#302](https://github.com/trusted-execution-clusters/operator/pull/302) for it. Once it gets merged, I will rebase this PR

tests/trusted_execution_cluster.rs
iroykaufman: The polling where the condition uses the pod logs can't be replaced with `await_condition`. There are two poller that doesn't use logs so I'll refactor them.

alicefr: we shouldn't relay on logs for validating the test. That also needs to be changed

Jakob-Naucke: I do get @iroykaufman's point here https://github.com/trusted-execution-clusters/operator/pull/248#discussion_r3492690952

tests/trusted_execution_cluster.rs
```diff
@@ -369,6 +372,424 @@ async fn test_nonexistent_approved_image() -> anyhow::Result<()> {
     let ctx = "waiting for ApprovedImage coreos1 to be PodPending";
     timeout(scaled_duration(30), done).await.context(ctx)??;
 
+     test_ctx.cleanup().await?;
+    Ok(())
+}
+}
+
+named_test! {
+async fn test_luks_key_sync() -> anyhow::Result<()> {
```
alicefr, line 381: Do we really need this test actually? We are already checking the attestation if a single VM and multiple, if the secrets and AKs weren't correctly propagate those test would fail. @Jakob-Naucke WDYT?

Jakob-Naucke: True. If we had some cascade test logic where tests that require more resources than unit tests (namely, a cluster) but fewer resources than an attestation test (a hypervisor) run first, this would be an interesting test case, but I wasn't planning on such infra anytime soon.

tests/trusted_execution_cluster.rs
alicefr: Instead of checking the logs we could exec in the pod and verify that the secret is gone if that's really necessary. It seems to me more reliable.

alicefr: Does trustee save the secrets and AKs in its filesystem since the API introduction?

tests/trusted_execution_cluster.rs
iroykaufman: The poller here is inside the `wait_for_deployment_ready` this is what you mean?

alicefr: yes, I would completely remove the `wait_for_deployment_ready`

alicefr:
@iroykaufman do you know where trustee stores the secrets and attestations keys after they have been configured via the API? I hope it stores it somewhere in its filesystem. If that's the case we need to persist them, otherwise if trustee restarts we are going to lose them

tests/trusted_execution_cluster.rs
iroykaufman: With the current kbs config the storage backend is configured to be on the filesystem.

uril:
@alicefr The current solution does not persist the trustee-data on the filesystem - when the Trustee pod restarts, the operator uploads all the keys (and other configuration) to the new pod.

Currently Trustee does keep those on its file-system, but that may change in the future. For example there is an idea of storing the data in a database.

tests/trusted_execution_cluster.rs
iroykaufman: Remove `wait_for_deployment_ready` and instead use `await_condition`? because I have to check that the object is ready.

Jakob-Naucke: Yes, à la https://github.com/trusted-execution-clusters/operator/blob/103fa3f9d10a697cf3ad2ad40543cc94fffce217/test_utils/src/lib.rs#L925 (which you could move up a level for reuse)

tests/trusted_execution_cluster.rs
Jakob-Naucke: Never mind, this test also restarts the Trustee deployment which is good to have. This ties back in to your other question on persistence: yes, we still have secrets and sync them, including on restarts (new function `trustee_deployment_reconcile`)

alicefr:
> > > @iroykaufman can you please also split the commit [774f99b](https://github.com/trusted-execution-clusters/operator/commit/774f99bba3467ab64ae03b0d75eb5a33f3534d2b) into multiple It does too many things in one. I would at least split the bump of the kbs-client version and the configuration into a separate commit.
> > 
> > 
> > @iroykaufman was the problem that the kbs-client won't work as used here with 17? if it does work, I agree. if it doesn't, I prefer big commits over breaking bisects @alicefr
> 
> Correct. When bumping to trustee v0.20.0, it didn't work as is because the trustee storage backend changed. The idea was to use the API as an abstraction layer that could make it much easier to bump trustee to a later version or change the storage backend in the future.

But is the old behavior of using config map for secrets, AK and RV still supported by 0.20? Or did they suddenly move and only support adding the various objects via API?

Jakob-Naucke:
> But is the old behavior of using config map for secrets, AK and RV still supported by 0.20? Or did they suddenly move and only support adding the various objects via API?

no, file system (NB it doesn't matter if it's a configmap) is no longer supported AFAICT, see https://github.com/confidential-containers/trustee/pull/1190

iroykaufman:
> > > > @iroykaufman can you please also split the commit [774f99b](https://github.com/trusted-execution-clusters/operator/commit/774f99bba3467ab64ae03b0d75eb5a33f3534d2b) into multiple It does too many things in one. I would at least split the bump of the kbs-client version and the configuration into a separate commit.
> > > 
> > > 
> > > @iroykaufman was the problem that the kbs-client won't work as used here with 17? if it does work, I agree. if it doesn't, I prefer big commits over breaking bisects @alicefr
> > 
> > 
> > Correct. When bumping to trustee v0.20.0, it didn't work as is because the trustee storage backend changed. The idea was to use the API as an abstraction layer that could make it much easier to bump trustee to a later version or change the storage backend in the future.
> 
> But is the old behavior of using config map for secrets, AK and RV still supported by 0.20? Or did they suddenly move and only support adding the various objects via API?

Trustee v0.20.0 replaced file-based configuration with a KV storage backend that Trustee manages internally it reads and writes to its own storage directories (e.g. `/opt/trustee/storage/kbs/`,` /opt/trustee/storage/reference_value/`). While it is still possible to mount these files with some extra effort, doing so in this version and moving forward does not appear to be the correct approach.

alicefr:
> > > > > @iroykaufman can you please also split the commit [774f99b](https://github.com/trusted-execution-clusters/operator/commit/774f99bba3467ab64ae03b0d75eb5a33f3534d2b) into multiple It does too many things in one. I would at least split the bump of the kbs-client version and the configuration into a separate commit.
> > > > 
> > > > 
> > > > @iroykaufman was the problem that the kbs-client won't work as used here with 17? if it does work, I agree. if it doesn't, I prefer big commits over breaking bisects @alicefr
> > > 
> > > 
> > > Correct. When bumping to trustee v0.20.0, it didn't work as is because the trustee storage backend changed. The idea was to use the API as an abstraction layer that could make it much easier to bump trustee to a later version or change the storage backend in the future.
> > 
> > 
> > But is the old behavior of using config map for secrets, AK and RV still supported by 0.20? Or did they suddenly move and only support adding the various objects via API?
> 
> Trustee v0.20.0 replaced file-based configuration with a KV storage backend that Trustee manages internally it reads and writes to its own storage directories (e.g. `/opt/trustee/storage/kbs/`,` /opt/trustee/storage/reference_value/`). While it is still possible to mount these files with some extra effort, doing so in this version and moving forward does not appear to be the correct approach.

I didn't ask because I want to keep the old behavior, but if the mounting of the secrets and other objects still works, we could gradually add each component one by one in separate commits. In this way, we keep the git bisect working but the PR becomes a bit more readable

iroykaufman:
> > > > > > @iroykaufman can you please also split the commit [774f99b](https://github.com/trusted-execution-clusters/operator/commit/774f99bba3467ab64ae03b0d75eb5a33f3534d2b) into multiple It does too many things in one. I would at least split the bump of the kbs-client version and the configuration into a separate commit.
> > > > > 
> > > > > 
> > > > > @iroykaufman was the problem that the kbs-client won't work as used here with 17? if it does work, I agree. if it doesn't, I prefer big commits over breaking bisects @alicefr
> > > > 
> > > > 
> > > > Correct. When bumping to trustee v0.20.0, it didn't work as is because the trustee storage backend changed. The idea was to use the API as an abstraction layer that could make it much easier to bump trustee to a later version or change the storage backend in the future.
> > > 
> > > 
> > > But is the old behavior of using config map for secrets, AK and RV still supported by 0.20? Or did they suddenly move and only support adding the various objects via API?
> > 
> > 
> > Trustee v0.20.0 replaced file-based configuration with a KV storage backend that Trustee manages internally it reads and writes to its own storage directories (e.g. `/opt/trustee/storage/kbs/`,` /opt/trustee/storage/reference_value/`). While it is still possible to mount these files with some extra effort, doing so in this version and moving forward does not appear to be the correct approach.
> 
> I didn't ask because I want to keep the old behavior, but if the mounting of the secrets and other objects still works, we could gradually add each component one by one in separate commits. In this way, we keep the git bisect working but the PR becomes a bit more readable

I understand, initially I tried to do exactly that, but the file formats also changed (you can choose to use either JSON or individual files). If I remember correctly, we use both formats: JSON for reference values and individual files for secrets. Because of this, it will create some extra work that will need to be reverted later. Maybe I missed things but this is what I stumbled upon.

uril:
Trustee v0.20.0 can still keep configuration files on its file-system, so it is possible to update to 0.20.0 first
As Roy mentioned it would take some adjustments of configuration files -- for example reference-values from json to files - each PCR has a separate file, instead of a single json file, and those adjustments will be overwritten by following patches.

tests/trusted_execution_cluster.rs
iroykaufman: @alicefr @Jakob-Naucke it is fine to just change wait_for_deployment_ready to use await_condition insted of polling?

alicefr:
@uril @iroykaufman thanks for the explication! No, I don't want you to do useless extra work. Let's keep it in this way then.

tests/trusted_execution_cluster.rs
Jakob-Naucke: yes, most preferably in a commit before the other ones, create the function and start using it from `apply_cr_manifests`, and then use it like you're already doing

uril:
@alicefr @iroykaufman we'll check again how much work it is to do the trustee upgrade to 0.20.0 before/after the API change. It would be nicer to have them in separate commits.

iroykaufman:
I want to summarize the comments and outline the actions I took:
- Integrate reflector: I started implementing this, but it turned out to be a relatively large change. This is because I need to forward the context from main to all the necessary functions. For example, `launch_rv_image_controller` and `launch_rv_job_controller` now need to accept `AkContextData` for `image_reconcile`, among other things.  I prefer to submit this change as a separate PR.
- Avoid log checks in the integration tests: I understand the motivation here is to prevent tests from failing if log messages change. However, the alternative approach, executing into the pod to check if resources are populated raises a similar issue and could actually be worse. If the file locations change, the test will still fail, but the added complexity will make it harder to debug and fix. With log checks, fixing a failure simply requires updating the expected log string.
-  All other comments are resolved.
As @uril mentioned, I'll try to do the trustee upgrade to v0.20.0 before/after the API change.
If there is anything I miss, let me know

alicefr:
> I want to summarize the comments and outline the actions I took:
> 
> * Integrate reflector: I started implementing this, but it turned out to be a relatively large change. This is because I need to forward the context from main to all the necessary functions. For example, `launch_rv_image_controller` and `launch_rv_job_controller` now need to accept `AkContextData` for `image_reconcile`, among other things.  I prefer to submit this change as a separate PR.

That's ok.

> * Avoid log checks in the integration tests: I understand the motivation here is to prevent tests from failing if log messages change. However, the alternative approach, executing into the pod to check if resources are populated raises a similar issue and could actually be worse. If the file locations change, the test will still fail, but the added complexity will make it harder to debug and fix. With log checks, fixing a failure simply requires updating the expected log string.

The log don't actually check the actual result but only that the part of the code was run. I still think checking the secrets in the filesystem will be better. Or as an alternative. you could run kbs-client from a container and query the object from the API and see that they are effectively gone. If this complicates the PR too much, we can leave the log and refactor it later. @Jakob-Naucke WDYT?

alicefr:
@iroykaufman for the rest, it isn't clear to me if this is ready for review or you want first try to split the trustee bump? thanks!

Jakob-Naucke:
> > * Avoid log checks in the integration tests: I understand the motivation here is to prevent tests from failing if log messages change. However, the alternative approach, executing into the pod to check if resources are populated raises a similar issue and could actually be worse. If the file locations change, the test will still fail, but the added complexity will make it harder to debug and fix. With log checks, fixing a failure simply requires updating the expected log string.
> 
> The log don't actually check the actual result but only that the part of the code was run. I still think checking the secrets in the filesystem will be better. Or as an alternative. you could run kbs-client from a container and query the object from the API and see that they are effectively gone. If this complicates the PR too much, we can leave the log and refactor it later. @Jakob-Naucke WDYT?

I too could see Trustee change the structure, so maybe kbs-client > logs > filesystem. I agree it can be a TODO & future PR

iroykaufman:
> > I want to summarize the comments and outline the actions I took:
> > 
> > * Integrate reflector: I started implementing this, but it turned out to be a relatively large change. This is because I need to forward the context from main to all the necessary functions. For example, `launch_rv_image_controller` and `launch_rv_job_controller` now need to accept `AkContextData` for `image_reconcile`, among other things.  I prefer to submit this change as a separate PR.
> 
> That's ok.
> 
> > * Avoid log checks in the integration tests: I understand the motivation here is to prevent tests from failing if log messages change. However, the alternative approach, executing into the pod to check if resources are populated raises a similar issue and could actually be worse. If the file locations change, the test will still fail, but the added complexity will make it harder to debug and fix. With log checks, fixing a failure simply requires updating the expected log string.
> 
> The log don't actually check the actual result but only that the part of the code was run. I still think checking the secrets in the filesystem will be better. Or as an alternative. you could run kbs-client from a container and query the object from the API and see that they are effectively gone. If this complicates the PR too much, we can leave the log and refactor it later. @Jakob-Naucke WDYT?

As I understand and from what I checked, the log is only printed when the API call is successful. For example, in `send_secret` there is a call to  `kbs_client::set_resource(..)` if the call was successful and trustee got the resource, the function will return ok and error otherwise. Only if the return value is ok the success log message will be printed.

iroykaufman:
> @iroykaufman for the rest, it isn't clear to me if this is ready for review or you want first try to split the trustee bump? thanks!

I'll give it a try today for splitting the commits. I'll let you know when it's ready

iroykaufman:
@alicefr @Jakob-Naucke, this is a status update on splitting the commits: I tried bumping `trustee` to v0.20.0 and using mount instead of the API calls. However, this introduces additional changes (not a very small amount because of the format change of the files) that will be deleted. I also tried keeping `trustee` on v0.17.0 and applying the three patches (such as the delete option) on `trustee`. This creates some merge conflicts on the trustee side that I will need to resolve. It is a bit challenging but doable. The main issue is that those commits will only work with this specific build of `trustee`. If you feel the current state of the PR is manageable to review, let me know. Otherwise, please don't hesitate to let me know, and I will implement one of the first two options, or if there is a different option, I'm happy to hear it.

On a separate note, I updated the whitelist of dependencies to the ones that came with `kbs-client` to the allowlist, but I am not sure how to verify if they are permitted by Fedora.

Jakob-Naucke:
> It is a bit challenging but doable. The main issue is that those commits will only work with this specific build of `trustee`. If you feel the current state of the PR is manageable to review, let me know.

It's a hard read but I think we could review as is. It's worth knowing that a split will be many more lines changed than its rollup (IIUC). @alicefr WDYT?

> On a separate note, I updated the whitelist of dependencies to the ones that came with `kbs-client` to the allowlist, but I am not sure how to verify if they are permitted by Fedora.

I'm afraid that many wouldn't be, but we can also fix this separately, as we are for oci-client. I'm sorry for insisting on adding the feature earlier -- I didn't expect this rabbit hole to be so deep.

alicefr:
> > It is a bit challenging but doable. The main issue is that those commits will only work with this specific build of `trustee`. If you feel the current state of the PR is manageable to review, let me know.
> 
> It's a hard read but I think we could review as is. It's worth knowing that a split will be many more lines changed than its rollup (IIUC). @alicefr WDYT?
> 
It doesn't make sense to overcomplicate this PR. We will review it as it is.

operator/src/trustee.rs
Jakob-Naucke: this is still open

operator/src/trustee.rs
Jakob-Naucke: indeed, we escalate at all other parts of the codebase. I think there's a potential argument for this more graceful approach but let's have it outside of this PR.

tests/trusted_execution_cluster.rs
```diff
@@ -368,6 +372,387 @@ async fn test_nonexistent_approved_image() -> anyhow::Result<()> {
...
+    // Wait for both K8s secrets to be created by the keygen controller
+    let secrets_api: Api<Secret> = Api::namespaced(client.clone(), namespace);
+    wait_for_resource_created(&secrets_api, &machine1_uuid, scaled_timeout(60)).await?;
+    wait_for_resource_created(&secrets_api, &machine2_uuid, scaled_timeout(60)).await?;
+    test_ctx.info("Both machine secrets created");
+
+    // Wait for the operator to send both secrets to the KBS
+    let pods_api: Api<Pod> = Api::namespaced(client.clone(), namespace);
+    let poller = Poller::new()
+        .with_timeout(Duration::from_secs(60))
```
Jakob-Naucke: still misses a scaled timeout here (and in a few other places, best to `git diff main | grep Duration` or sth like that

tests/trusted_execution_cluster.rs
Jakob-Naucke: can `apply_cr_manifests` use the new function?

Jakob-Naucke:
oh, and could you pull https://github.com/iroykaufman/operator/pull/1?

operator/src/trustee.rs
iroykaufman: This comment is on outdated code, so I want to mention that it has been changed to return requeue for 30 seconds. This is necessary because the reconcile function can be called before the trustee address is updated.

operator/src/trustee.rs
iroykaufman: Using the reflector requires propagating the context across all functions, so it is better suited for a separate PR. See the comment [here](https://github.com/trusted-execution-clusters/operator/pull/248#issuecomment-4968665926)

iroykaufman:
> oh, and could you pull [iroykaufman#1](https://github.com/iroykaufman/operator/pull/1)?

Sure, but why is this required on OpenShift SCC?

iroykaufman:
Uri and I found a way to split the commits without adding more complexity and keep it working on every commit. While it isn't a major split, it should make it more readable. I've put the changes in [PR#2](https://github.com/iroykaufman/operator/pull/2) on my fork. The final diff is the same as this one, aside from a few placement differences. @alicefr @Jakob-Naucke Let me know if you prefer this approach and would like me to push it here. Sorry for the back and forth.

Jakob-Naucke:
> Uri and I found a way to split the commits

yes, pull it in, it seems a lot more readable. thank you!

> > oh, and could you pull [iroykaufman#1](https://github.com/iroykaufman/operator/pull/1)?
> 
> Sure, but why is this required on OpenShift SCC?

In OpenShift, under the `restricted-v2` SCC, writes to `/opt/trustee/storage` fail because the container's UID doesn't have permission. Sorry for putting so little comment in the PR.

tests/trusted_execution_cluster.rs
```diff
@@ -368,6 +372,387 @@ async fn test_nonexistent_approved_image() -> anyhow::Result<()> {
...
+                let lp = ListParams::default().labels("app=trusted-cluster-operator");
+                let operator_pods = api.list(&lp).await?;
+                let pod_name = operator_pods
+                    .items
+                    .first()
+                    .and_then(|p| p.metadata.name.as_ref())
+                    .ok_or_else(|| anyhow::anyhow!("Operator pod not found"))?
+                    .clone();
+                let logs = api.logs(&pod_name, &LogParams::default()).await?;
+                if logs.contains("Syncing 2 machine luks key to KBS") {
```
Jakob-Naucke: This caused a CI failure (after many successful runs):
```
2026-07-30T15:15:59Z INFO: test_luks_key_sync: Trustee deployment is ready after restart
2026-07-30T15:16:01Z INFO: test_luks_key_sync: Both secrets re-synced to KBS after trustee restart
2026-07-30T15:16:01Z INFO: test_luks_key_sync: Deleted Machine 1: test-machine-84805090
Error: Machine1 secret not deleted from KBS, last error was: Machine1 secret not yet deleted from KBS
FAILED
```

operator:
```
[2026-07-30T15:16:00Z INFO  operator::trustee] Sending secret 84805090-056a-42ea-b3ea-453c76d47098 to KBS API...
[2026-07-30T15:16:00Z INFO  operator::trustee] Secret 84805090-056a-42ea-b3ea-453c76d47098 sent successfully
...
[2026-07-30T15:16:01Z INFO  operator::trustee] Deleting secret 84805090-056a-42ea-b3ea-453c76d47098 to KBS API...
[2026-07-30T15:16:02Z ERROR operator] failed to reconcile on machine: failed to clean up object: failed to clean up object: failed to delete secret for machine 84805090-056a-42ea-b3ea-453c76d47098: Request Failed, Response: "{\"type\":\"https://github.com/confidential-containers/kbs/errors/PluginInternalError\",\"detail\":\"Plugin internal error\"}"
[2026-07-30T15:16:02Z INFO  operator] reconcile failed: ReconcilerFailed(Anyhow(failed to reconcile on machine: failed to clean up object: failed to clean up object: failed to delete secret for machine 84805090-056a-42ea-b3ea-453c76d47098: Request Failed, Response: "{\"type\":\"https://github.com/confidential-containers/kbs/errors/PluginInternalError\",\"detail\":\"Plugin internal error\"}"), ObjectRef { dyntype: ApiResource { group: "trusted-execution-clusters.io", version: "v1alpha1", api_version: "trusted-execution-clusters.io/v1alpha1", kind: "Machine", plural: "machines" }, name: "test-machine-84805090", namespace: Some("test-9c457981"), extra: Extra { resource_version: Some("100782"), uid: Some("976e67a5-fed9-4e28-bc63-ab2e5bd2f139") } })
```

trustee:
```
$ head -1 trustee-deployment-test-9c457981.log
2026-07-30T15:16:00.255706Z  INFO kbs: Welcome to Confidential Containers Key Broker Service!
$ grep 84805090 trustee-deployment-test-9c457981.log
2026-07-30T15:16:02.762283Z ERROR kbs::error: PluginInternalError { source: resource not found: default/84805090-056a-42ea-b3ea-453c76d47098/root }
2026-07-30T15:16:02.762298Z DEBUG actix_web::middleware::logger: Error in response: PluginInternalError { source: resource not found: default/84805090-056a-42ea-b3ea-453c76d47098/root }
2026-07-30T15:16:02.762308Z  INFO actix_web::middleware::logger: 10.244.82.32 "DELETE /kbs/v0/resource/default/84805090-056a-42ea-b3ea-453c76d47098/root HTTP/1.1" 401 117 "-" "kbs-client/0.1.0" 0.000684
2026-07-30T15:16:02.945496Z  INFO actix_web::middleware::logger: 10.244.82.32 "POST /kbs/v0/resource/default/84805090-056a-42ea-b3ea-453c76d47098/root HTTP/1.1" 200 0 "-" "kbs-client/0.1.0" 0.000571
```

so my theory is that Trustee's deployment was ready but it wasn't handling requests properly yet (despite answering with HTTP OK), so it also had no secret to delete. Technically that's a bug in Trustee, but before filing that we should bump to latest which is 0.22 by now, which is maybe out of scope for this PR. My hunch would be to put a 2s wait on the resync and a TODO comment, but it's admittedly dirty.

SpaceFace02:
Just a little side-note which was raised in another PR:
https://github.com/trusted-execution-clusters/operator/pull/324#discussion_r3688874886

iroykaufman:
> Just a little side-note which was raised in another PR: [#324 (comment)](https://github.com/trusted-execution-clusters/operator/pull/324#discussion_r3688874886)

Every time there is a change in one of the AK (e.g one is deleted), the AK finalizer is triggered, and then `trustee::update_attestation_keys` is called and overwrites the reference value `trusted_aks`  with the updated list of AKs (e.g if one is deleted, the list will not include this AK). @SpaceFace02 let me know if this answers your comment

SpaceFace02:
> Every time there is a change in one of the AK (e.g one is deleted), the AK finalizer is triggered, and then trustee::update_attestation_keys is called and overwrites the reference value trusted_aks with the updated list of AKs (e.g if one is deleted, the list will not include this AK)

https://github.com/trusted-execution-clusters/operator/pull/324/changes#diff-baaf5539b868bb9602782a44f6ff07c54184e08f63cc0e44c5931ff34f832d22R219

Yes, but my minor nit was basically in this comment, which says `"Machine {} is being deleted, updating attestation key volumes",`

Because there are no more attestation key volumes, this comment is a bit misleading, that's all.

iroykaufman:
> > Every time there is a change in one of the AK (e.g one is deleted), the AK finalizer is triggered, and then trustee::update_attestation_keys is called and overwrites the reference value trusted_aks with the updated list of AKs (e.g if one is deleted, the list will not include this AK)
> 
> https://github.com/trusted-execution-clusters/operator/pull/324/changes#diff-baaf5539b868bb9602782a44f6ff07c54184e08f63cc0e44c5931ff34f832d22R219
> 
> Yes, but my minor nit was basically in this comment, which says `"Machine {} is being deleted, updating attestation key volumes",`
> 
> Because there are no more attestation key volumes, this comment is a bit misleading, that's all.

Right, I'll change this

tests/trusted_execution_cluster.rs
iroykaufman: I'm trying to reproduce this bug, but it is not working for me. How many runs until this happens? I may not have run it enough times. 

I think the problem may be related to the fact that the check of the log message only checks that the sync of the secret begins and not that it succeeds. So, maybe this sends the delete request before the resource is sent successfully (Although I'm seeing in the log messages that the secret sends successfully, but there is a need to have two of those messages). To solve this problem, we check as follows:
```rust
if logs.contains("Syncing 2 machine luks key to KBS")
   && logs.matches(&format!("{id1} sent successfully")).count() >= 2
   && logs.matches(&format!("{id2} sent successfully")).count()  >= 2
```

Jakob-Naucke: Right, that's also a possible, and maybe more plausible explanation. Could you add this check?

iroykaufman: Yes, I'm on it

Jakob-Naucke: > How many runs until this happens?

should have been after ~10 although that's not exactly an MTBF statistic

Jakob-Naucke (approved):
nit: can you commit this change to Cargo.lock where you change Cargo.toml

```diff
diff --git a/Cargo.lock b/Cargo.lock
index d71482b..cac9319 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1215 +1215 @@ dependencies = [
- "thiserror 2.0.18",
+ "thiserror 2.0.19",
```

iroykaufman:
@Jakob-Naucke do you know what went wrong with the CI? In the first attempt, I got this error
```
FAILED
Error: SSH access to VM did not become available after 600 seconds, last error was: SSH not available yet: ssh command failed: You are using a client virtctl version that is different from the KubeVirt version running in the cluster
Client Version: v1.8.4
Server Version: v1.9.0

Internal error occurred: dialing VM: dial tcp 10.244.82.44:22: connect: connection refused
Connection closed by UNKNOWN port 65535
You are using a client virtctl version that is different from the KubeVirt version running in the cluster
Client Version: v1.8.4
Server Version: v1.9.0

exit status 255

```
And after this attempt, the job just canceled in the middle of the run

Jakob-Naucke:
@iroykaufman first failure is potentially remediated in the follow-up to #324, second I have no idea :( re-running

operator/src/main.rs
```diff
@@ -289,9 +296,9 @@ async fn main() -> Result<()> {
     attestation_key_register::launch_ak_controller(ak_ctx.clone()).await;
     attestation_key_register::launch_machine_ak_controller(ak_ctx.clone()).await;
     attestation_key_register::launch_secret_ak_controller(ak_ctx).await;
-    reference_values::create_pcrs_config_map(kube_client.clone()).await?;
```
alicefr, line 292: why moving this?

operator/src/trustee.rs
```diff
@@ -308,213 +321,111 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
-        .secret_store
-        .state()
-        .into_iter()
+pub async fn update_attestation_keys(client: Client) -> Result<()> {
+    let secrets: Api<Secret> = Api::default_namespaced(client.clone());
+    let secret_list = secrets.list(&Default::default()).await?;
+
+    let ak_secrets: Vec<String> = secret_list
+        .items
+        .iter()
```
alicefr, line 405: Can we repserve the original code that was already using the cache, or any reasons why doing direct api call?

operator/src/trustee.rs
```diff
@@ -93,24 +106,211 @@ fn recompute_reference_values(image_pcrs: ImagePcrs) -> Vec<ReferenceValue> {
...
+            tec.spec
+                .public_trustee_addr
+                .as_ref()
+                .context("TrustedExecutionCluster missing public_trustee_addr HTTP")?
+        ),
+        vec![],
+    ))
+}
+
+async fn sync_reference_values(client: &Client, reference_values: &[ReferenceValue]) -> Result<()> {
```
Jakob-Naucke, line 190: I realize this is late to ask, but does this support removal of reference values?

operator/src/trustee.rs
iroykaufman: This function is now called within the trustee deployment reconcile, so changing it to accept the cache context will cause more changes in other functions that will have to accept the cache context. I already started working on a PR it's currently under my fork [here](https://github.com/iroykaufman/operator/pull/3). I think it would be best to introduce this after this PR gets merged.

Jakob-Naucke: I think it's fine to add in a follow-up

operator/src/trustee.rs
Jakob-Naucke: as discussed synchronously, it's overwrite-all

operator/src/main.rs
iroykaufman: This was unnecessary. I probably changed it while trying to solve a problem and forgot to revert it, so I have deleted the change.

iroykaufman:
Sorry for the noise from all the pushes some changes I made got mixed up. Hopefully this one is good.

operator/src/trustee.rs
```diff
@@ -105,24 +116,143 @@ fn recompute_reference_values(image_pcrs: ImagePcrs) -> Vec<ReferenceValue> {
...
+                .public_trustee_addr
+                .as_ref()
+                .context("TrustedExecutionCluster missing public_trustee_addr HTTPS")?
+        );
+        return Ok((trustee_addr, vec![ca_pem]));
+    }
+
+    Ok((
+        format!(
+            "http://{}",
```
alicefr, line 174: as a TODO for next PR, this requires to use the tls certificate from cert manager

operator/src/trustee.rs
```diff
@@ -316,213 +329,111 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
-            mount_path: format!("{TRUSTEE_SECRETS_PATH}/{id}"),
-            ..Default::default()
-        },
-    )
+pub fn secret_path(id: &str) -> String {
+    format!("default/{id}/root")
+}
+
+pub async fn send_secret(client: Client, id: &str) -> Result<()> {
+    let secret_api: Api<Secret> = Api::default_namespaced(client.clone());
```
alicefr, line 337: As a TODO for next step, we need to use the cache here

operator/src/trustee.rs
```diff
@@ -316,213 +329,111 @@ fn generate_luks_key() -> Result<Vec<u8>> {
...
 }
 
-pub async fn update_attestation_keys(ctx: &AkContextData) -> Result<()> {
-    let client = &ctx.client;
-    let ak_secrets: Vec<String> = ctx
-        .secret_store
-        .state()
-        .into_iter()
+pub async fn update_attestation_keys(client: Client) -> Result<()> {
+    let secrets: Api<Secret> = Api::default_namespaced(client.clone());
```
alicefr, line 408: Same here it needs to use the reflector, as next TODO

tests/trusted_execution_cluster.rs
```diff
@@ -285,6 +289,392 @@ async fn test_nonexistent_approved_image() -> anyhow::Result<()> {
...
+                let operator_pods = api.list(&lp).await?;
+                let pod_name = operator_pods
+                    .items
+                    .first()
+                    .and_then(|p| p.metadata.name.as_ref())
+                    .ok_or_else(|| anyhow::anyhow!("Operator pod not found"))?
+                    .clone();
+                let logs = api.logs(&pod_name, &LogParams::default()).await?;
+                if logs.contains(&format!("{id1} sent successfully"))
+                    && logs.contains(&format!("{id2} sent successfully"))
```
alicefr, line 375: As next TODO, we need to check this properly with exec and not using the logs

alicefr (approved):

Issue #258, opened 2026-05-19
Replace kube API patch and merge directives with Server side Apply directives


SpaceFace02:
Currently in [this update_status](https://github.com/trusted-execution-clusters/operator/blob/f14b212178437ee9a5d94b19482ff5c0a829bf6d/lib/src/lib.rs#L33) macro, we are making use of kube client PATCH directive which does a client side diffing and updates the CRD. 

There are other places where we've used client side directives, but those are single owner scenarious ([here](https://github.com/trusted-execution-clusters/operator/blob/f14b212178437ee9a5d94b19482ff5c0a829bf6d/operator/src/attestation_key_register.rs#L254) and [here](https://github.com/trusted-execution-clusters/operator/blob/f14b212178437ee9a5d94b19482ff5c0a829bf6d/operator/src/reference_values.rs#L293))

We have used SSA (Server side Apply) [here](https://github.com/trusted-execution-clusters/operator/blob/f14b212178437ee9a5d94b19482ff5c0a829bf6d/operator/src/trustee.rs#L317), which prevents race conditions better than client side Patch or Merge.

SSA introduces a managedFields field in the CRD, which ensures that controllers only update the sections they are responsible for. This prevents concurrent overwrites and last-write-win conditions.

references:
1. https://clustersandcoffee.substack.com/p/the-beauty-of-kubernetes-server-side
2. https://kubernetes.io/docs/reference/using-api/server-side-apply/

PR #260, opened 2026-05-20
feat: if condition exists, updating it, otherwise inserting it to the conditions list in the CRD, in the reconcile function

SpaceFace02:
Bug fix for: https://github.com/trusted-execution-clusters/operator/issues/220

Instead of recreating the conditions from scratch each time the reconcile function runs, this makes sure that foreign controller conditions are not overwritten, and same type conditions are updated, rather then recreated; new conditions are inserted.

## Summary by Sourcery

Preserve and incrementally update status conditions during reconcile instead of rebuilding them from scratch, while tightening related tests and minor code ergonomics.

Bug Fixes:
- Ensure reconcile upserts status conditions so operator-owned conditions are updated or inserted without overwriting conditions from other controllers.
- Fix PCR computation handling by correctly tracking when PCRs should be recomputed for new images.

Enhancements:
- Introduce a shared upsert_condition helper to update or append conditions by type.
- Refine TLS startup logic for services to use safer option pattern matching.
- Rename the register server service port to a more specific name to avoid ambiguity in service configuration.
- Improve test helpers to assert multiple expected substrings in PATCH bodies with clearer error messages.
- Clarify documentation for running individual integration tests and required environment setup.

Tests:
- Extend reconcile tests to verify that foreign controller conditions are preserved while operator conditions are inserted or updated, and that only a single Installed condition is present and transitioned to True.
- Update existing reconcile tests to use the new multi-string assertion helper for PATCH request validation.

### DIFF
--- a/attestation-key-register/src/main.rs
+++ b/attestation-key-register/src/main.rs
@@ -138,9 +138,8 @@ async fn main() {
     let service = app.into_make_service();
     info!("Starting attestation key registration server on http://{addr}",);
 
-    let run = if args.cert_path.is_some() && args.key_path.is_some() {
-        let config = OpenSSLConfig::from_pem_file(args.cert_path.unwrap(), args.key_path.unwrap())
-            .expect("invalid PEM files");
+    let run = if let (Some(cert_path), Some(key_path)) = (args.cert_path, args.key_path) {
+        let config = OpenSSLConfig::from_pem_file(cert_path, key_path).expect("invalid PEM files");
         axum_server::bind_openssl(addr, config).serve(service).await
     } else {
         axum_server::bind(addr).serve(service).await
--- a/operator/src/main.rs
+++ b/operator/src/main.rs
@@ -16,7 +16,7 @@ use kube::runtime::watcher;
 use kube::{Api, Client};
 use log::{error, info, warn};
 
-use operator::generate_owner_reference;
+use operator::{generate_owner_reference, upsert_condition};
 use trusted_cluster_operator_lib::{TrustedExecutionCluster, TrustedExecutionClusterStatus};
 use trusted_cluster_operator_lib::{conditions::*, update_status};
 
@@ -107,7 +107,11 @@ async fn reconcile(
     let existing_status = &cluster.status;
     let address_condition =
         known_trustee_address_condition(known_address, generation, existing_status);
-    let mut conditions = Some(vec![address_condition]);
+
+    // Get existing conditions or default to empty vector
+    let mut conditions = existing_status.as_ref().and_then(|s| s.conditions.clone());
+    // Update or insert address condition to prevent rebuilding the status object from scratch every time the reconcile is called.
+    let _ = upsert_condition(&mut conditions, address_condition);
 
     let kube_client = ctx.client.clone();
     let err = "trusted execution cluster had no name";
@@ -117,9 +121,12 @@ async fn reconcile(
     if cluster.metadata.deletion_timestamp.is_some() {
         info!("Registered deletion of TrustedExecutionCluster {name}");
         let uninstalling_reason = NOT_INSTALLED_REASON_UNINSTALLING;
-        let condition = installed_condition(uninstalling_reason, generation, existing_status);
-        conditions.as_mut().unwrap().push(condition);
-        update_status!(clusters, name, TrustedExecutionClusterStatus { conditions })?;
+        let uninstall_condition =
+            installed_condition(uninstalling_reason, generation, existing_status);
+        let changed = upsert_condition(&mut conditions, uninstall_condition);
+        if changed {
+            update_status!(clusters, name, TrustedExecutionClusterStatus { conditions })?;
+        }
         return Ok(Action::await_change());
     }
 
@@ -137,29 +144,35 @@ async fn reconcile(
             "More than one TrustedExecutionCluster found in namespace {namespace}. \
              trusted-cluster-operator does not support more than one TrustedExecutionCluster. Requeueing...",
         );
-        let condition =
+        let non_unique_condition =
             installed_condition(NOT_INSTALLED_REASON_NON_UNIQUE, generation, existing_status);
-        conditions.as_mut().unwrap().push(condition);
-        update_status!(clusters, name, TrustedExecutionClusterStatus { conditions })?;
+        let changed = upsert_condition(&mut conditions, non_unique_condition);
+        if changed {
+            update_status!(clusters, name, TrustedExecutionClusterStatus { conditions })?;
+        }
         return Ok(Action::requeue(Duration::from_secs(60)));
     }
 
     info!("Setting up TrustedExecutionCluster {name}");
-    let mut installing = conditions.clone();
-    let installing_reason = NOT_INSTALLED_REASON_INSTALLING;
-    let condition = installed_condition(installing_reason, generation, existing_status);
-    installing.as_mut().unwrap().push(condition);
-    let status = TrustedExecutionClusterStatus {
-        conditions: installing,
-    };
-    update_status!(clusters, name, status)?;
+    let installing_condition =
+        installed_condition(NOT_INSTALLED_REASON_INSTALLING, generation, existing_status);
+    let changed = upsert_condition(&mut conditions, installing_condition);
+    if changed {
+        let status = TrustedExecutionClusterStatus {
+            conditions: conditions.clone(),
+        };
+        update_status!(clusters, name, status)?;
+    }
 
     install_trustee_configuration(kube_client.clone(), &cluster).await?;
     install_register_server(kube_client.clone(), &cluster).await?;
     install_attestation_key_register(kube_client, &cluster).await?;
-    let condition = installed_condition(INSTALLED_REASON, generation, existing_status);
-    conditions.as_mut().unwrap().push(condition);
-    update_status!(clusters, name, TrustedExecutionClusterStatus { conditions })?;
+    let installed_condition = installed_condition(INSTALLED_REASON, generation, existing_status);
+    let changed = upsert_condition(&mut conditions, installed_condition);
+    if changed {
+        let status = TrustedExecutionClusterStatus { conditions };
+        update_status!(clusters, name, status)?;
+    }
     Ok(Action::await_change())
 }
 
@@ -358,7 +371,8 @@ mod tests {
     async fn test_reconcile_uninstalling() {
         let clos = async |req: Request<Body>, ctr| match req.method() {
             &Method::PATCH => {
-                assert_body_contains(req, NOT_INSTALLED_REASON_UNINSTALLING).await;
+                let body = get_body_string(req).await;
+                assert!(body.contains(NOT_INSTALLED_REASON_UNINSTALLING),);
                 Ok(serde_json::to_string(&dummy_cluster()).unwrap())
             }
             _ => panic!("unexpected API interaction: {req:?}, counter {ctr}"),
@@ -385,7 +399,8 @@ mod tests {
                 // Watchers
                 Ok(serde_json::to_string(&dummy_cluster()).unwrap())
             } else if ctr == 4 && req.method() == Method::PATCH {
-                assert_body_contains(req, NOT_INSTALLED_REASON_NON_UNIQUE).await;
+                let body = get_body_string(req).await;
+                assert!(body.contains(NOT_INSTALLED_REASON_NON_UNIQUE));
                 Ok(serde_json::to_string(&dummy_cluster()).unwrap())
             } else {
                 panic!("unexpected API interaction: {req:?}, counter {ctr}");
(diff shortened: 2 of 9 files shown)


SpaceFace02:
I believe we have a setting which makes ```cargo fmt``` and ```clippy``` return warnings as errors, which made me change a few other parts of the code I did not modify for the feature. Is this intentional?

lib/src/lib.rs
```diff
@@ -153,3 +153,13 @@ pub async fn get_trusted_execution_cluster(
 
     Ok(list.items.pop().unwrap())
 }
```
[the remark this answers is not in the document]
SpaceFace02: @alicefr @Jakob-Naucke 

Out of the 3 suggestions, I find 2 to be valid (first and last)

1. the upsert function is identifying based on type. Maybe we need another field called Owner (part of this is mentioned in the issue https://github.com/trusted-execution-clusters/operator/issues/258
2. conditions is cloned as we are using conditions even after the installing status update, to update final status, so reconcile needs to maintain ownership.
3. Bug risk: Does seem valid, WDYT?

alicefr:
> I believe we have a setting which makes `cargo fmt` and `clippy` return warnings as errors, which made me change a few other parts of the code I did not modify for the feature. Is this intentional?

Weird, we have this [job](https://github.com/trusted-execution-clusters/operator/blob/main/.github/workflows/lint.yml) that check the cargo and clippy. Might be that the version you have installed is more recent than the toolchain in the CI

EDIT: from the CI the version is `toolchain install 1.88.0 `

lib/src/lib.rs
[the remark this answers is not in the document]
alicefr: > Bug risk: Does seem valid, WDYT?

@SpaceFace02 this seems very close to https://github.com/trusted-execution-clusters/operator/issues/220, It was found by analyzing the code with Claude

operator/src/reference_values.rs
```diff
@@ -366,7 +366,9 @@ pub async fn handle_new_image(
         return Ok(NOT_COMMITTED_REASON_NO_DIGEST);
     }
     let label = fetch_pcr_label(&image_ref).await;
-    let compute_pcrs = match label {
+
+    // Whether to compute pcrs or not.
+    let should_compute_pcrs = match label {
```
alicefr, line 376: nit: any particular reason to change the variable name?

operator/src/register_server.rs
```diff
@@ -108,7 +108,8 @@ pub async fn create_register_server_service(
         spec: Some(ServiceSpec {
             selector: Some(labels),
             ports: Some(vec![ServicePort {
-                name: Some("http".to_string()),
+                // Renamed to register-server-port for disambiguation from http, as the presence of a secret indicates http or https.
+                name: Some("register-server-port".to_string()),
```
alicefr, line 111: can you move it in another commit?

tests/README.md
```diff
@@ -23,6 +23,8 @@ make integration-tests
 ```
 
 Each test can also be run independently using cargo test. Example:
+
+Note: before running independent tests, make sure TRUSTEE_IMAGE, APPROVED_IMAGE, TEST_IMAGE, RUST_LOG, REGISTRY, TAG env variables are set. Also run `make trusted-cluster-gen` to create the minimal/default boiler-plate yaml manifests for the TEC.
```
alicefr: nit: can you split this also in a separate commit

alicefr:
@SpaceFace02 thanks, just a couple of nits

lib/src/lib.rs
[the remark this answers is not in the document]
alicefr: we can fix the bug as part of 220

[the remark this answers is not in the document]
alicefr: > the upsert function is identifying based on type. Maybe we need another field called Owner (part of this is mentioned in the issue https://github.com/trusted-execution-clusters/operator/issues/258

let's introduce when necessary. Right now that's not the case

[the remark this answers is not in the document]
SpaceFace02: This PR attempts to fix #220 only, have picked it as a good first issue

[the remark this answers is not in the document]
alicefr: you are right I somehow confused it with https://github.com/trusted-execution-clusters/operator/issues/221.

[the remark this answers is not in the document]
alicefr: I think the bug is valid. You are always overwriting the condition and its last_transition_time. The condition needs to be updated only if the status changed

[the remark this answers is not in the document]
SpaceFace02: The last_transition_time value will remain same, because the transition_time [helper](https://github.com/trusted-execution-clusters/operator/blob/main/lib/src/lib.rs#L70), makes sure that the timestamp doesn't change if condition hasnt changed.

But yes, talking about performance, it does have unnecessary patches which might not be needed. So I'll update the condition only if any of the field has changed.

operator/src/reference_values.rs
SpaceFace02: It was a nitpick from my side too while going through the codebase, just to rename the variable name for clarity, initially it was named ```compute_pcrs```, which could also suggest that these are the computed_pcrs, but ```should_compute_pcrs``` implies a boolean

lib/src/lib.rs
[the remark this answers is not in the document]
alicefr: can you add a third test where the condition doesn't need to be updated and each field is preserved, including the last_transition_time. I guess this should spot the bug discovered by sourcery

attestation-key-register/src/main.rs
```diff
@@ -138,9 +138,8 @@ async fn main() {
     let service = app.into_make_service();
     info!("Starting attestation key registration server on http://{addr}",);
 
-    let run = if args.cert_path.is_some() && args.key_path.is_some() {
-        let config = OpenSSLConfig::from_pem_file(args.cert_path.unwrap(), args.key_path.unwrap())
-            .expect("invalid PEM files");
+    let run = if let (Some(cert_path), Some(key_path)) = (args.cert_path, args.key_path) {
```
Jakob-Naucke, line 141: (also in response to

>> I believe we have a setting which makes cargo fmt and clippy return warnings as errors, which made me change a few other parts of the code I did not modify for the feature. Is this intentional?
>
> Weird, we have this [job](https://github.com/trusted-execution-clusters/operator/blob/main/.github/workflows/lint.yml) that check the cargo and clippy. Might be that the version you have installed is more recent than the toolchain in the CI

)

I agree with this change, and you're invited to keep making them (in separate commits like you did here), but let me give some context as to why it's not caught by CI in the first place: This has only become a lint recently, and we use a Rust version supported by RHEL/UBI to avoid downstream patching when using a Rust feature that's newer than that. We also use the linting as of that version because sometimes a new Rust feature also becomes a lint to use that feature relatively fast or even in the same version.

*With that said*, we could update to 1.92 (not for this PR). Ideally it's even possible to get dependabot to update to a version supported by e.g. `registry.access.redhat.com/ubi9`. This could maybe also go in a doc then.

tests/README.md
Jakob-Naucke: good addition, but nit for flow & formatting
```suggestion
Each test can also be run independently using cargo test.
Before running independent tests, run `make trusted-cluster-gen` to create the default boiler-plate yaml manifests for the TEC and make sure `TRUSTEE_IMAGE`, `APPROVED_IMAGE`, `TEST_IMAGE`, `RUST_LOG`, `REGISTRY`, `TAG` env variables are set.

Example:
```

operator/src/register_server.rs
```diff
@@ -108,7 +108,8 @@ pub async fn create_register_server_service(
         spec: Some(ServiceSpec {
             selector: Some(labels),
             ports: Some(vec![ServicePort {
-                name: Some("http".to_string()),
+                // Renamed to register-server-port for disambiguation from http, as the presence of a secret indicates http or https.
```
Jakob-Naucke: good change, but this comment should be the commit message -- no need to have the code history in the code itself

lib/src/lib.rs
```diff
@@ -153,3 +153,24 @@ pub async fn get_trusted_execution_cluster(
...
+
+// Update condition if already present, otherwise append(insert) it into the conditions vector.
+pub fn upsert_condition(conditions: &mut Option<Vec<Condition>>, condition: Condition) -> bool {
```
Jakob-Naucke: I guess this is a bit of a rewrite of https://github.com/kubernetes/apimachinery/blob/a7225b106d19985c2a47654c106cb419007af6d2/pkg/api/meta/conditions.go#L31. Not for this PR, but you could add a TODO to contribute that functionality to kube-rs.

On a different note, this should go to `operator/src/lib.rs` because you're only using it in-operator.

test_utils/src/mock_client.rs
```diff
@@ -107,10 +107,17 @@ where
     }
 }
 
-pub async fn assert_body_contains(req: Request<Body>, contains: &str) {
+pub async fn assert_body_contains(req: Request<Body>, expected: &[(&str, Option<&str>)]) {
     let bytes = req.into_body().collect_bytes().await.unwrap().to_vec();
     let body = String::from_utf8_lossy(&bytes);
-    assert!(body.contains(contains));
+
+    for (expected_reason, expected_err_msg) in expected {
```
Jakob-Naucke: I feel the loop refactor is a little overkill for one use

operator/src/main.rs
```diff
@@ -410,4 +435,188 @@ mod tests {
...
+
+    // This test ensures that if the condition is not changed, the status is not patched. The transition_time and all other fields remain same.
+    #[tokio::test]
+    async fn test_reconcile_no_patch_when_conditions_unchanged() {
+        let clos1 = async |req: Request<Body>, _| match *req.method() {
+            Method::PATCH => Ok(serde_json::to_string(&dummy_cluster()).unwrap()),
+            _ => panic!("unexpected: {req:?}"),
+        };
+
+        {
```
Jakob-Naucke: this check, and the one below, don't need to be in blocks

operator/src/main.rs
```diff
@@ -410,4 +435,188 @@ mod tests {
             assert!(result.is_err());
         });
     }
+
+    // Makes sure that uninstall trigger preserves foreign independent controller conditions, and our operator doesn't overwrite it in the reconcile function. Tests insert of our upsert_condition function.
+    #[tokio::test]
+    async fn test_reconcile_uninstall_preserves_foreign_controller_condition_by_inserting_owned_condition()
+     {
+        let foreign_condition = Condition {
```
Jakob-Naucke: but I do think this Condition could be generated from a dummy function

operator/src/main.rs
```diff
@@ -410,4 +435,188 @@ mod tests {
             assert!(result.is_err());
         });
     }
+
+    // Makes sure that uninstall trigger preserves foreign independent controller conditions, and our operator doesn't overwrite it in the reconcile function. Tests insert of our upsert_condition function.
+    #[tokio::test]
+    async fn test_reconcile_uninstall_preserves_foreign_controller_condition_by_inserting_owned_condition()
```
Jakob-Naucke: Which owned condition? Maybe just `test_reconcile_uninstall_preserves_foreign_condition`. Same for `test_reconcile_install_preserves_foreign_condition_while_updating_owned_condition`.

operator/src/main.rs
```diff
@@ -410,4 +435,188 @@ mod tests {
...
+                Method::POST => Ok(serde_json::to_string(&dummy_cluster()).unwrap()),
+                Method::PATCH => {
+                    let body = req.into_body().collect_bytes().await.unwrap().to_vec();
+                    let body = String::from_utf8_lossy(&body);
+                    assert!(
+                        body.contains("ForeignCondition"),
+                        "Request body did not contain ForeignCondition, got: {body}"
+                    );
+
+                    if body.contains(INSTALLED_REASON) {
```
Jakob-Naucke: Why is this in a block, don't we always expect that there is exactly one Installed condition?

Jakob-Naucke (commented):
- `upsert_condition` should also be used by `image_add_reconcile`
- Please put fixes by `rustfmt` in the same commit

Tip: If you put `Fixes: #220` in a commit message, it will auto-link that issue. I've linked it for now.

lib/src/lib.rs
SpaceFace02: That's a great find, yes I'll add the TODO to port it to kube-rs in the future.

And yes, I'll move it to ```operator/src/lib.rs```

test_utils/src/mock_client.rs
SpaceFace02: Yes, I didn't want to do it initially, but in order to test multiple conditions and to see whether all fields are preserved, I added this. 

this method needs ownership of body, as it consumes it, and I didn't want to clone too many times in the caller function, so I thought this is a better way to handle this, albeit a bit verbose.

LMK

operator/src/main.rs
SpaceFace02: as these tests belong to operator, this is related to all conditions owned by operator for managing the TEC CRD. In the future, we is possible that we may have certain fields of TEC not owned by operator, and owned by some foreign controller

operator/src/main.rs
SpaceFace02: Yes, there is only 1 installed condition, but 2 patch calls with different reasons (one with Installing reason and one for InstallationCompleted reason).

Here I am checking that in the last PATCH call (with body containing Reason=InstallationCompleted), does reconcile update existing Installed Condition with new status and reason (expected behaviour), or does it append another condition instead (what we don't want).

operator/src/main.rs
SpaceFace02: Yes, good catch, I am using it in 2 places with exact initialization

operator/src/main.rs
SpaceFace02: Yes, I had added blocks because there was some import scoping issue of atomic crate, it was being imported in the macro, and since it wasn't scoped initially, it was throwing a double import issue, have fixed it.

SpaceFace02:
> upsert_condition should also be used by image_add_reconcile

Have added the upsert_condition for both image_add_reconcile and ak_reconcile.

SpaceFace02:
> Tip: If you put Fixes: #220 in a commit message, it will auto-link that issue. I've linked it for now.

By adding this to commit message, it will reference the commit instead of the PR right, so once that commit checks in the main branch, will it close the PR and resolve the issue?

image

operator/src/main.rs
Jakob-Naucke: Ah, right. In that case: nit: maybe do an early return here, the indenting is already really deep

test_utils/src/mock_client.rs
Jakob-Naucke: Right, I see the problem, the Body isn't `Clone` because it could be a `Wrap`… honestly I'd just change the `assert_body_contains` function to something like

```rust
pub async fn get_body_string(req: Request<Body>) -> String {
    let bytes = req.into_body().collect_bytes().await.unwrap().to_vec();
    String::from_utf8_lossy(&bytes).to_string()
}
```

and assert on that from the callers, saves a lot of lines:

```diff
diff --git a/operator/src/main.rs b/operator/src/main.rs
index 05b4868..682ca03 100644
--- a/operator/src/main.rs
+++ b/operator/src/main.rs
@@ -371,15 +371,8 @@ mod tests {
     async fn test_reconcile_uninstalling() {
         let clos = async |req: Request<Body>, ctr| match req.method() {
             &Method::PATCH => {
-                let uninstall_assertion_reason = "Request body did not contain Uninstall condition";
-                assert_body_contains(
-                    req,
-                    &[(
-                        NOT_INSTALLED_REASON_UNINSTALLING,
-                        Some(uninstall_assertion_reason),
-                    )],
-                )
-                .await;
+                let body = get_body_string(req).await;
+                assert!(body.contains(NOT_INSTALLED_REASON_UNINSTALLING));
                 Ok(serde_json::to_string(&dummy_cluster()).unwrap())
             }
             _ => panic!("unexpected API interaction: {req:?}, counter {ctr}"),
@@ -406,9 +399,8 @@ mod tests {
                 // Watchers
                 Ok(serde_json::to_string(&dummy_cluster()).unwrap())
             } else if ctr == 4 && req.method() == Method::PATCH {
-                let err_msg = "Request body did not contain Non-Unique condition";
-                assert_body_contains(req, &[(NOT_INSTALLED_REASON_NON_UNIQUE, Some(err_msg))])
-                    .await;
+                let body = get_body_string(req).await;
+                assert!(body.contains(NOT_INSTALLED_REASON_NON_UNIQUE));
                 Ok(serde_json::to_string(&dummy_cluster()).unwrap())
             } else {
                 panic!("unexpected API interaction: {req:?}, counter {ctr}");
@@ -453,24 +445,10 @@ mod tests {
 
         let clos = async |req: Request<Body>, ctr| match req.method() {
             &Method::PATCH => {
-                assert_body_contains(
-                    req,
-                    &[
-                        (
-                            "ForeignCondition",
-                            Some("Request body did not contain ForeignCondition"),
-                        ),
-                        (
-                            "ExternalController",
-                            Some("Request body did not contain ExternalController"),
-                        ),
-                        (
-                            NOT_INSTALLED_REASON_UNINSTALLING,
-                            Some("Request body did not contain Uninstall condition"),
-                        ),
-                    ],
-                )
-                .await;
+                let body = get_body_string(req).await;
+                assert!(body.contains("ForeignCondition"));
+                assert!(body.contains("ExternalController"));
+                assert!(body.contains(NOT_INSTALLED_REASON_UNINSTALLING));
                 Ok(serde_json::to_string(&dummy_cluster()).unwrap())
             }
             _ => panic!("unexpected API interaction: {req:?}, counter {ctr}"),
diff --git a/test_utils/src/mock_client.rs b/test_utils/src/mock_client.rs
index aa0a259..08e34e8 100644
--- a/test_utils/src/mock_client.rs
+++ b/test_utils/src/mock_client.rs
@@ -106,17 +106,9 @@ where
     }
 }
 
-pub async fn assert_body_contains(req: Request<Body>, expected: &[(&str, Option<&str>)]) {
+pub async fn get_body_string(req: Request<Body>) -> String {
     let bytes = req.into_body().collect_bytes().await.unwrap().to_vec();
-    let body = String::from_utf8_lossy(&bytes);
-
-    for (expected_reason, expected_err_msg) in expected {
-        let msg = expected_err_msg.unwrap_or("Given body doesn't contain the expected string");
-        assert!(
-            body.contains(expected_reason),
-            "{msg}: expected '{expected_reason}', got: {body}"
-        );
-    }
+    String::from_utf8_lossy(&bytes).to_string()
 }
 
 pub async fn test_create_success<
```

Jakob-Naucke (commented):
> > Tip: If you put Fixes: #220 in a commit message, it will auto-link that issue. I've linked it for now.
> 
> By adding this to commit message, it will reference the commit instead of the PR right, so once that commit checks in the main branch, will it close the PR and resolve the issue?

The issue gets auto-resolved when a PR that was linked to it is merged, and one way to link the PR to the issue is to add a commit to the PR that references this issue with a `Fixes:` tag.

Another nit on the commit message: Try to keep the subject line short and put the rest of the description in the body (ideally, subject line is at most 50 characters, but that's not a hard and fast rule)

operator/src/main.rs
```diff
@@ -453,24 +450,19 @@ mod tests {
...
-                            NOT_INSTALLED_REASON_UNINSTALLING,
-                            Some("Request body did not contain Uninstall condition"),
-                        ),
-                    ],
-                )
-                .await;
+                let body = get_body_string(req).await;
+                assert!(
+                    body.contains("ForeignCondition"),
+                    "Request body did not contain ForeignCondition"
```
Jakob-Naucke: nit: I think the increased lines of adding assertion messages (+3 because of formatting) distract more from reading the test code than they help understanding a failure, the failure will always contain a line number

test_utils/src/mock_client.rs
```diff
@@ -106,17 +106,15 @@ where
     }
 }
 
-pub async fn assert_body_contains(req: Request<Body>, expected: &[(&str, Option<&str>)]) {
+pub async fn assert_body_contains(req: Request<Body>, contains: &str) {
     let bytes = req.into_body().collect_bytes().await.unwrap().to_vec();
```
Jakob-Naucke: this could be refactored to use `get_body_string`, or you drop the function as a whole

Jakob-Naucke (commented):
and please amend the updates into the respective commits :) for the failed tests, GH seems to have one of its moments but they're hopefully away until your next push

SpaceFace02:
> and please amend the updates into the respective commits :) for the failed tests, GH seems to have one of its moments but they're hopefully away until your next push

Yeah, looks like GH Actions had a temporary hiccup — I just re-pushed, and it seems to be working now

Jakob-Naucke (approved):

PR #267, opened 2026-06-01
Introduce the reflector to reduce the k8s API calls

### LINKED ISSUE #251: Introduce the reflector to reduce the k8s API calls
Currently, our operator performs several direct calls to the Kubernetes API server to get and list various objects. Each direct call to the API server incurs network latency and consumes resources on the API server and etcd. Additionally, performing frequent reads against the API server risks hitting rate limits (HTTP 429) during heavy workloads.

Replace direct read operations (get and list) with the kube-rs Reflector. By establishing a background Watcher that keeps an in-memory cache of the resources, we can serve read requests locally.

Example of directl calls:
+ [machine list in the AK reconcile loop](https://github.com/trusted-execution-clusters/operator/blob/main/operator/src/attestation_key_register.rs#L141)
+ [ak list in the machine reconcile loop](https://github.com/trusted-execution-clusters/operator/blob/main/operator/src/attestation_key_register.rs#L175)
+ [get_trusted_execution_cluster](https://github.com/trusted-execution-clusters/operator/blob/main/lib/src/lib.rs#L85)
+ [get of the config map](https://github.com/trusted-execution-clusters/operator/blob/main/operator/src/trustee.rs#L96)
+ [get of the deployment](https://github.com/trusted-execution-clusters/operator/blob/main/operator/src/trustee.rs#L156)

yairpod:
As requested in [issue 251](https://github.com/trusted-execution-clusters/operator/issues/251).
This PR replaces the direct k8s API calls with reflectors.
This should reduce the API server network latency and consumption of resources on the API server and etcd as well as  risks hitting rate limits (HTTP 429) during heavy workloads.
The cost is increased memory usage as we cache the information from the API server (the CPU for the watcher threads should be reasonably low).
I have not been able to test the cache memory usage as a kind cluster will have a tiny memory signature anyway.

To test the API call reduction I have run the Integration tests on a clean kind cluster and run the command `kubectl get --raw /metrics  | grep apiserver_request_total | grep trusted`, then summed the all calls of all types.
The unmodified operator had a total of 788 calls.
The PR code had a total of 463 calls.

## Summary by Sourcery

Introduce shared reflector-backed caches for custom resources and core objects to reduce direct Kubernetes API usage and wire them into the operator controllers.

New Features:
- Add reflector-based in-memory stores for TrustedExecutionCluster, Machine, AttestationKey, Secret, and Deployment resources and wait for their initial synchronization before starting controllers.

Enhancements:
- Refactor attestation key controllers to share a common context carrying Kubernetes client and reflector stores instead of repeatedly listing resources from the API server.
- Use the reflector-backed TrustedExecutionCluster store in the main reconcile path to detect non-unique clusters without performing a fresh list call.
- Update trustee attestation key volume management to derive secrets and deployment state from reflector caches rather than live API reads.
- Adjust and extend tests to work with pre-populated reflector stores and reduced API interactions, while keeping existing behavior coverage.

### DIFF
--- a/operator/src/attestation_key_register.rs
+++ b/operator/src/attestation_key_register.rs
@@ -15,12 +15,19 @@ use k8s_openapi::apimachinery::pkg::{
 };
 use kube::{
     Api, Client, Resource,
-    api::{ListParams, ObjectList, Patch, PatchParams},
-    runtime::{Controller, controller::Action, finalizer, finalizer::Event, watcher},
+    api::{Patch, PatchParams},
+    runtime::{
+        Controller,
+        controller::Action,
+        finalizer,
+        finalizer::Event,
+        reflector::{self, ObjectRef, Store},
+        watcher,
+    },
 };
 use log::info;
 use serde_json::json;
-use std::{collections::BTreeMap, sync::Arc};
+use std::{collections::BTreeMap, sync::Arc, time::Duration};
 
 use trusted_cluster_operator_lib::conditions::ATTESTATION_KEY_MACHINE_APPROVE;
 use trusted_cluster_operator_lib::endpoints::*;
@@ -32,6 +39,46 @@ use operator::{
     ControllerError, TLS_DIR, controller_error_policy, create_or_info_if_exists, read_certificate,
     upsert_condition,
 };
+
+/// Shared context for the three attestation-key controllers.
+/// Stores give local cache access to avoid repeated API-server reads.
+pub struct AkContextData {
+    pub client: Client,
+    pub machine_store: Store<Machine>,
+    pub ak_store: Store<AttestationKey>,
+    pub secret_store: Store<Secret>,
+    pub deployment_store: Store<Deployment>,
+}
+
+impl AkContextData {
+    pub fn new(client: Client) -> Self {
+        let (machine_store, machine_writer) = reflector::store::<Machine>();
+        let (ak_store, ak_writer) = reflector::store::<AttestationKey>();
+        let (secret_store, secret_writer) = reflector::store::<Secret>();
+        let (deployment_store, deployment_writer) = reflector::store::<Deployment>();
+
+        crate::spawn_reflector::<Machine>(machine_writer, client.clone(), "Machine");
+        crate::spawn_reflector::<AttestationKey>(ak_writer, client.clone(), "AttestationKey");
+        crate::spawn_reflector::<Secret>(secret_writer, client.clone(), "Secret");
+        crate::spawn_reflector::<Deployment>(deployment_writer, client.clone(), "Deployment");
+
+        Self {
+            client,
+            machine_store,
+            ak_store,
+            secret_store,
+            deployment_store,
+        }
+    }
+
+    pub async fn sync_caches(&self, timeout: Duration) -> Result<()> {
+        crate::sync_cache(&self.machine_store, "Machine", timeout).await?;
+        crate::sync_cache(&self.ak_store, "AttestationKey", timeout).await?;
+        crate::sync_cache(&self.secret_store, "Secret", timeout).await?;
+        crate::sync_cache(&self.deployment_store, "Deployment", timeout).await?;
+        Ok(())
+    }
+}
 
 const INTERNAL_ATTESTATION_KEY_REGISTER_PORT: i32 = 8001;
 const ATTESTATION_KEY_SECRET_FINALIZER: &str =
@@ -140,21 +187,14 @@ pub async fn create_attestation_key_register_service(
 
 async fn ak_reconcile(
     ak: Arc<AttestationKey>,
-    client: Arc<Client>,
+    ctx: Arc<AkContextData>,
 ) -> Result<Action, ControllerError> {
     let ak_name = ak.metadata.name.clone().unwrap_or_default();
     info!("Attestation Key reconciliation for: {ak_name}");
 
-    let client = Arc::unwrap_or_clone(client);
-    let machines: Api<Machine> = Api::default_namespaced(client.clone());
-    let lp = ListParams::default();
-    let machine_list: ObjectList<Machine> = machines.list(&lp).await.map_err(|e| {
-        eprintln!("Error fetching machine list: {e}");
-        ControllerError::Anyhow(e.into())
-    })?;
-    for machine in &machine_list.items {
+    for machine in ctx.machine_store.state() {
         if ak.spec.uuid.as_ref() == Some(&machine.spec.id) {
-            approve_ak(&ak, machine, client.clone()).await?;
+            approve_ak(&ak, &machine, &ctx).await?;
             return Ok(Action::await_change());
         }
     }
@@ -163,13 +203,12 @@ async fn ak_reconcile(
 
 async fn machine_reconcile(
     machine: Arc<Machine>,
-    client: Arc<Client>,
+    ctx: Arc<AkContextData>,
 ) -> Result<Action, ControllerError> {
     info!(
         "Machine reconciliation for: {}",
         machine.metadata.name.clone().unwrap_or_default()
     );
-    let client = Arc::unwrap_or_clone(client);
 
     // Check if the machine is being deleted
     if machine.metadata.deletion_timestamp.is_some() {
@@ -180,25 +219,20 @@ async fn machine_reconcile(
         return Ok(Action::await_change());
     }
 
-    let aks: Api<AttestationKey> = Api::default_namespaced(client.clone());
-    let lp = ListParams::default();
-    let ak_list: ObjectList<AttestationKey> = aks.list(&lp).await.map_err(|e| {
-        eprintln!("Error fetching attestation key list: {e}");
-        ControllerError::Anyhow(e.into())
-    })?;
-    for ak in ak_list.items {
+    for ak in ctx.ak_store.state() {
         if let Some(ak_uuid) = &ak.spec.uuid
             && *ak_uuid == machine.spec.id
         {
-            approve_ak(&ak, &machine, client.clone()).await?;
+            approve_ak(&ak, &machine, &ctx).await?;
             return Ok(Action::await_change());
         }
     }
     Ok(Action::await_change())
 }
 
-async fn approve_ak(ak: &AttestationKey, machine: &Machine, client: Client) -> Result<()> {
+async fn approve_ak(ak: &AttestationKey, machine: &Machine, ctx: &AkContextData) -> Result<()> {
     let name = ak.metadata.name.clone().unwrap_or_default();
+    let client = &ctx.client;
     let aks: Api<AttestationKey> = Api::default_namespaced(client.clone());
 
     let generation = ak.metadata.generation;
@@ -241,8 +275,11 @@ async fn approve_ak(ak: &AttestationKey, machine: &Machine, client: Client) -> R
     }
 
     let secret_name = name.clone();
-    let secrets: Api<Secret> = Api::default_namespaced(client.clone());
-    let secret_exists = secrets.get(&secret_name).await.is_ok();
+    let ns = client.default_namespace().to_string();
+    let secret_exists = ctx
+        .secret_store
+        .get(&ObjectRef::new(&secret_name).within(&ns))
+        .is_some();
 
     if !secret_exists {
         let public_key_data = ByteString(ak.spec.public_key.as_bytes().to_vec());
@@ -270,7 +307,7 @@ async fn approve_ak(ak: &AttestationKey, machine: &Machine, client: Client) -> R
 
 async fn secret_reconcile(
     secret: Arc<Secret>,
-    client: Arc<Client>,
+    ctx: Arc<AkContextData>,
 ) -> Result<Action, ControllerError> {
     let secret_name = secret.metadata.name.clone().unwrap_or_default();
 
@@ -288,13 +325,13 @@ async fn secret_reconcile(
 
     info!("Secret reconciliation for AttestationKey secret: {secret_name}");
 
-    let secrets: Api<Secret> = Api::default_namespaced(Arc::unwrap_or_clone(client.clone()));
+    let secrets: Api<Secret> = Api::default_namespaced(ctx.client.clone());
+    let ctx = ctx.clone();
     finalizer(&secrets, ATTESTATION_KEY_SECRET_FINALIZER, secret, |ev| async move {
         match ev {
             Event::Apply(_secret) => {
                 // On creation/update, just update the trustee deployment volumes
-                let client = Arc::unwrap_or_clone(client);
-                trustee::update_attestation_keys(client)
+                trustee::update_attestation_keys(&ctx)
                     .await
                     .map(|_| Action::await_change())
                     .map_err(|e| {
@@ -307,21 +344,16 @@ async fn secret_reconcile(
                 info!(
                     "AttestationKey secret {secret_name} is being deleted, updating trustee deployment volumes"
                 );
-                let client = Arc::unwrap_or_clone(client);
                 // Update trustee deployment - secrets with deletion_timestamp will be filtered out
-                match trustee::update_attestation_keys(client).await {
-                    Ok(_) => Ok(Action::await_change()),
-                    Err(e) if e.to_string().contains("not found") => {
-                        info!("Trustee deployment not found during secret cleanup (likely already deleted)");
-                        Ok(Action::await_change())
-                    }
-                    Err(e) => {
+                trustee::update_attestation_keys(&ctx)
+                    .await
+                    .map(|_| Action::await_change())
+                    .map_err(|e| {
                         eprintln!(
                             "Error updating attestation key volumes during secret deletion: {e}"
                         );
-                        Err(finalizer::Error::<ControllerError>::CleanupFailed(e.into()))
-                    }
-                }
+                        finalizer::Error::<ControllerError>::CleanupFailed(e.into())
+                    })
             }
         }
     })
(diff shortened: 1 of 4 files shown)


operator/src/main.rs
```diff
@@ -295,18 +297,97 @@ async fn main() -> Result<()> {
     env_logger::Builder::from_env(Env::default().default_filter_or("info")).init();
 
     let kube_client = Client::try_default().await?;
-    info!("trusted execution clusters operator",);
+    info!("trusted execution clusters operator");
+
+    // --- Create in-memory stores (Store + Writer pairs) ---
```
Jakob-Naucke: nit: Please follow other comments' styling (no `--- … ---`)

operator/src/main.rs
```diff
@@ -330,9 +411,22 @@ mod tests {
         ClusterContext {
             client,
             uid: Mutex::new(None),
+            tec_store: reflector::store::<TrustedExecutionCluster>().0,
         }
     }
 
+    /// Build a Store pre-populated with two distinct TrustedExecutionCluster objects.
```
Jakob-Naucke, line 315: 2 TEC objects are unsupported. Why do you want to do this?

operator/src/attestation_key_register.rs
```diff
@@ -33,6 +40,16 @@ use operator::{
...
+/// Shared context for the three attestation-key controllers.
+/// Stores give local cache access to avoid repeated API-server reads.
+pub struct AkContextData {
+    pub client: Client,
+    pub machine_store: Store<Machine>,
+    pub ak_store: Store<AttestationKey>,
+    pub secret_store: Store<Secret>,
+    pub deployment_store: Store<Deployment>,
+}
+
```
alicefr, line 52: @Jakob-Naucke I'm wondering if the AkContentData should implement the functions that use the stores. In this way we avoid to pass the store as argument. WDYT?

operator/src/main.rs
```diff
@@ -295,18 +299,97 @@ async fn main() -> Result<()> {
...
+    );
+
+    // --- Wait for stores to be populated from the initial List ---
+    info!("Waiting for caches to sync...");
+    tec_store.wait_until_ready().await?;
+    machine_store.wait_until_ready().await?;
+    ak_store.wait_until_ready().await?;
+    secret_store.wait_until_ready().await?;
+    deployment_store.wait_until_ready().await?;
+    info!("Caches synced, starting controllers");
```
alicefr: why not incapsulating these steps in an init fuctions, maybe even the new function for the type ak_ctx

operator/src/attestation_key_register.rs
Jakob-Naucke: but then you're still passing `AkContextData`s?

Jakob-Naucke (commented):
- don't worry about the CI failure, quay.io is having trouble. That said, ensure your code has been run through formatting (e.g. `cargo fmt`)
- is this also possible for `ApprovedImages` (`reference_values.rs:331`)?
- commit message nits: use infinitive (`Replace direct kube`…) remove `Subject: `

operator/src/attestation_key_register.rs
alicefr: Nevermind, it would also change the signature of the reconcile functions which makes the code more complicated. It was for grouping the functions using the same context

operator/src/main.rs
yairpod: Sounds good, added to new version

Jakob-Naucke: can't it be a function instead of a macro?
e: ah no, I see
e2: no, it can be, see comment

operator/src/main.rs
yairpod: It's used for the test_reconcile_non_unique test to inject a second TEC to the test.

operator/src/main.rs
yairpod: Removed

operator/src/main.rs
yairpod: Moved to a init function

operator/src/attestation_key_register.rs
```diff
@@ -140,21 +222,14 @@ pub async fn create_attestation_key_register_service(
...
-    let lp = ListParams::default();
-    let machine_list: ObjectList<Machine> = machines.list(&lp).await.map_err(|e| {
-        eprintln!("Error fetching machine list: {e}");
-        ControllerError::Anyhow(e.into())
-    })?;
-    for machine in &machine_list.items {
+    for machine in ctx.machine_store.state() {
         if ak.spec.uuid.as_ref() == Some(&machine.spec.id) {
-            approve_ak(&ak, machine, client.clone()).await?;
+            approve_ak(&ak, &machine, ctx.client.clone(), &ctx.secret_store).await?;
```
Jakob-Naucke: I think by that point you could just pass a reference to the context

operator/src/trustee.rs
```diff
@@ -190,13 +191,14 @@ pub async fn do_mount_secret(client: Client, id: &str, add: bool) -> Result<()>
...
-    let secret_list = secrets.list(&Default::default()).await?;
-
-    let ak_secrets: Vec<String> = secret_list
-        .items
-        .iter()
+pub async fn update_attestation_keys(
+    client: Client,
+    secret_store: &Store<Secret>,
+    deployment_store: &Store<Deployment>,
+) -> Result<()> {
```
Jakob-Naucke: Also just take a ctx?

operator/src/main.rs
Jakob-Naucke: yes, but why is that something you want to test?

operator/src/main.rs
```diff
@@ -295,19 +339,27 @@ async fn main() -> Result<()> {
...
+            .await?,
+    );
+    let ctx = Arc::new(ClusterContext::new(kube_client.clone(), CACHE_SYNC_TIMEOUT).await?);
+
+    info!("Caches synced, starting controllers");
+
     let cl: Api<TrustedExecutionCluster> = Api::default_namespaced(kube_client.clone());
 
-    // Launch all controllers except reference value-related ones
+    //  Launch all controllers
```
Jakob-Naucke: why this diff?

operator/src/main.rs
```diff
@@ -47,6 +50,49 @@ struct ClusterContext {
...
+                ),
+            )
+            .for_each(|res| async {
+                if let Err(e) = res {
+                    warn!("TrustedExecutionCluster reflector error: {e}");
+                }
+            }),
+        );
+
+        timeout(sync_timeout, tec_store.wait_until_ready())
```
Jakob-Naucke: Shouldn't this use the same sync_cache! macro?

operator/src/attestation_key_register.rs
```diff
@@ -33,6 +41,80 @@ use operator::{
...
+                );
+            };
+        }
+
+        spawn_reflector!(machine_writer, Machine, "Machine");
+        spawn_reflector!(ak_writer, AttestationKey, "AttestationKey");
+        spawn_reflector!(secret_writer, Secret, "Secret");
+        spawn_reflector!(deployment_writer, Deployment, "Deployment");
+
+        macro_rules! sync_cache {
```
Jakob-Naucke: can turn into a function for less confusing error messages if you ever change something:
```rust
async fn sync_cache<K>(store: &Store<K>, name: &str, sync_timeout: Duration) -> Result<()>
where
    K: 'static + Clone + reflector::Lookup,
    K::DynamicType: Eq + std::hash::Hash + Clone,
{
    let err = anyhow!(
        "Timed out after {sync_timeout:?} waiting for {name} cache to sync. \
         Ensure the CRD is installed and the API server is reachable."
    );
    timeout(sync_timeout, store.wait_until_ready())
        .await
        .map_err(|_| err)?
        .map_err(|e| anyhow!("Cache writer for {name} was dropped: {e}"))?;
    Ok(())
}
```

operator/src/attestation_key_register.rs
```diff
@@ -33,6 +41,80 @@ use operator::{
...
+    /// Create a fully-synced context: spawns background reflectors for all four
+    /// resource types and waits for their initial list-watch round-trip to
+    /// complete before returning.
+    pub async fn new(client: Client, sync_timeout: Duration) -> Result<Self> {
+        let (machine_store, machine_writer) = reflector::store::<Machine>();
+        let (ak_store, ak_writer) = reflector::store::<AttestationKey>();
+        let (secret_store, secret_writer) = reflector::store::<Secret>();
+        let (deployment_store, deployment_writer) = reflector::store::<Deployment>();
+
+        macro_rules! spawn_reflector {
```
Jakob-Naucke: can also be a function:
```rust
fn spawn_reflector<K>(writer: reflector::store::Writer<K>, client: Client, name: &'static str)
where
    K: Resource<Scope = k8s_openapi::NamespaceResourceScope>,
    K: Clone + serde::de::DeserializeOwned + std::fmt::Debug + Send + Sync + 'static,
    K::DynamicType: Default + Eq + std::hash::Hash + Clone,
{
    let watcher = watcher(Api::<K>::default_namespaced(client), Default::default());
    let reflector = reflector::reflector(writer, watcher).for_each(move |res| async move {
        if let Err(e) = res {
            warn!("{name} reflector error: {e}");
        }
    });
    tokio::spawn(reflector);
}
```

Jakob-Naucke (commented):
still wondering if it can also be implemented for `ApprovedImages` (`reference_values.rs:331`)?

operator/src/main.rs
yairpod: The test checks that attempting to install a second TEC results in a graceful denial with the correct error code "NOT_INSTALLED_REASON_NON_UNIQUE", so it needs to simulate that condition to the operator (see line 403 of the original code and 458 of the PR code).
If you think the test is unnecessary we can remove it, but might want to do that in a separate commit/PR.

Jakob-Naucke: ah of course, my bad

alicefr:
@yairpod try to use Opus as model instead of Sonnet. It will give you much better result

operator/src/trustee.rs
yairpod: Done

operator/src/main.rs
yairpod: Done

operator/src/attestation_key_register.rs
yairpod: Done

operator/src/attestation_key_register.rs
yairpod: Done

operator/src/main.rs
```diff
@@ -29,8 +32,40 @@ mod test_utils;
 mod trustee;
 
 use crate::conditions::*;
+use kube::Resource;
 use operator::*;
 
+fn spawn_reflector<K>(writer: reflector::store::Writer<K>, client: Client, name: &'static str)
```
Jakob-Naucke: I think these two should be in `operator/src/lib.rs`

operator/src/attestation_key_register.rs
```diff
@@ -288,52 +325,59 @@ async fn secret_reconcile(
...
-                    Err(e) if e.to_string().contains("not found") => {
-                        info!("Trustee deployment not found during secret cleanup (likely already deleted)");
-                        Ok(Action::await_change())
+    let secrets: Api<Secret> = Api::default_namespaced(ctx.client.clone());
+    let ctx = ctx.clone();
+    finalizer(
+        &secrets,
+        ATTESTATION_KEY_SECRET_FINALIZER,
+        secret,
+        move |ev| {
```
Jakob-Naucke: This reformat isn't required by rustfmt afaict, please keep the diff minimal

Jakob-Naucke (commented):
Can the reflector also be used for reference values?

operator/src/main.rs
yairpod: Moved

operator/src/attestation_key_register.rs
yairpod: Fixed

Jakob-Naucke: No, I mean like
```diff
     let ctx = ctx.clone();
-    finalizer(&secrets, ATTESTATION_KEY_SECRET_FINALIZER, secret, move |ev| {
-            let ctx = ctx.clone();
-            async move {
-                match ev {
-                    Event::Apply(_secret) => {
-                        // On creation/update, just update the trustee deployment volumes
-                        trustee::update_attestation_keys(&ctx)
-                            .await
-                            .map(|_| Action::await_change())
-                            .map_err(|e| {
-                                eprintln!("Error updating attestation key volumes on secret apply: {e}");
-                                finalizer::Error::<ControllerError>::ApplyFailed(e.into())
-                            })
-                    }
-                    Event::Cleanup(secret) => {
-                        let secret_name = secret.metadata.name.clone().unwrap_or_default();
-                        info!(
-                            "AttestationKey secret {secret_name} is being deleted, updating trustee deployment volumes"
+    finalizer(&secrets, ATTESTATION_KEY_SECRET_FINALIZER, secret, |ev| async move {
+        match ev {
+            Event::Apply(_secret) => {
+                // On creation/update, just update the trustee deployment volumes
+                trustee::update_attestation_keys(&ctx)
+                    .await
+                    .map(|_| Action::await_change())
+                    .map_err(|e| {
+                        eprintln!("Error updating attestation key volumes on secret apply: {e}");
+                        finalizer::Error::<ControllerError>::ApplyFailed(e.into())
+                    })
+            }
+            Event::Cleanup(secret) => {
+                let secret_name = secret.metadata.name.clone().unwrap_or_default();
+                info!(
+                    "AttestationKey secret {secret_name} is being deleted, updating trustee deployment volumes"
+                );
+                // Update trustee deployment - secrets with deletion_timestamp will be filtered out
+                trustee::update_attestation_keys(&ctx)
+                    .await
+                    .map(|_| Action::await_change())
+                    .map_err(|e| {
+                        eprintln!(
+                            "Error updating attestation key volumes during secret deletion: {e}"
                         );
-                        // Update trustee deployment - secrets with deletion_timestamp will be filtered out
-                        trustee::update_attestation_keys(&ctx)
-                            .await
-                            .map(|_| Action::await_change())
-                            .map_err(|e| {
-                                eprintln!(
-                                    "Error updating attestation key volumes during secret deletion: {e}"
-                                );
-                                finalizer::Error::<ControllerError>::CleanupFailed(e.into())
-                            })
-                    }
-                }
+                        finalizer::Error::<ControllerError>::CleanupFailed(e.into())
+                    })
             }
-        },
-    )
+        }
+    })
     .
```

and then your diff to main is much smaller & reviewable

operator/src/attestation_key_register.rs
```diff
@@ -140,21 +187,14 @@ pub async fn create_attestation_key_register_service(
...
 
-    let client = Arc::unwrap_or_clone(client);
-    let machines: Api<Machine> = Api::default_namespaced(client.clone());
-    let lp = ListParams::default();
-    let machine_list: ObjectList<Machine> = machines.list(&lp).await.map_err(|e| {
-        eprintln!("Error fetching machine list: {e}");
-        ControllerError::Anyhow(e.into())
-    })?;
-    for machine in &machine_list.items {
+    for machine in ctx.machine_store.state() {
```
Jakob-Naucke, line 195: This is where I'm not sure if we're adding fragility: This function reconciles on AKs, and we're replacing a list of Machines. If the cache is out of date, and eventually updates, is this reconciliation retriggered?

Jakob-Naucke (commented):
nit: use infinitive in the commit message subject (`Replace …`)

operator/src/attestation_key_register.rs
Jakob-Naucke: it should be possible to keep `|ev| async move {` on one line like it is in HEAD

operator/src/attestation_key_register.rs
yairpod: I think that in that case machine_reconcile will be called when the machine does arrive, and it will call approve_ak.

operator/src/attestation_key_register.rs
yairpod: Change applied

Jakob-Naucke: ah yes, gorgeous

operator/src/attestation_key_register.rs
Jakob-Naucke: ah of course, great

Jakob-Naucke (approved):
LGTM, would still like @alicefr's opinion

operator/src/main.rs
```diff
@@ -149,8 +176,11 @@ async fn install_trustee_configuration(
         Err(e) => error!("Failed to create the KBS service: {e}"),
     }
 
-    let default = format!("{TEC_REGISTRY}/key-broker-service:{TRUSTEE_VERSION}");
-    let trustee_image = env::var(RELATED_IMAGE_TRUSTEE).ok().unwrap_or(default);
+    let trustee_image = get_image_or_env(
```
Jakob-Naucke: rebase fault

operator/src/lib.rs
```diff
@@ -89,6 +94,37 @@ pub async fn read_certificate(
...
+
+pub async fn sync_cache<K>(store: &Store<K>, name: &str, sync_timeout: Duration) -> Result<()>
+where
+    K: 'static + Clone + reflector::Lookup,
+    K::DynamicType: Eq + std::hash::Hash + Clone,
+{
+    let err = anyhow::anyhow!(
+        "Timed out after {sync_timeout:?} waiting for {name} cache to sync. \
+         Ensure the CRD is installed and the API server is reachable."
+    );
```
alicefr, line 120: if this can be caused by the CRD to not be installed we can error out earlier by getting the CRD. It will also verify that the API server is reachable or not. We can improve this in a follow up PR though.

operator/src/main.rs
```diff
@@ -31,12 +32,40 @@ mod trustee;
...
-const TEC_REGISTRY: &str = "quay.io/trusted-execution-clusters";
+
+/// Get image from CR spec, falling back to environment variable (set by OLM), then default.
+/// This enables disconnected/airgap installations where OLM rewrites RELATED_IMAGE_* env vars.
+fn get_image_or_env(cr_image: &Option<String>, env_var: &str, default: &str) -> String {
+    cr_image
+        .clone()
+        .or_else(|| env::var(env_var).ok())
+        .unwrap_or_else(|| default.to_string())
+}
```
alicefr: Is this really par of this PR? Or is it still a rebase issue

yairpod: Rebase issue, solved

operator/src/lib.rs
yairpod: Sounds good for a follow up improvement

alicefr (approved):

PR #269, opened 2026-06-02
feat: migrate to Server-Side Apply (SSA) and enforce controller owner…

### LINKED ISSUE #258: Replace kube API patch and merge directives with Server side Apply directives

SpaceFace02:
feat: migrate to Server-Side Apply (SSA) and enforce controller ownership
Fixes: #258

Migrate resource creation and updates from client-side POST/PUT to Server-Side Apply (SSA) where appropriate, and strengthen ownership semantics with controller references throughout.

This commit establishes ownership of various fields and CRD's, to prevent accidental overwrites or race conditions by other controllers It doesn't migrate all patches to SSA, if we plan to migrate completely to SSA, there would be significant code rework, and honestly, its not necessary for now.

All resources created by the operator now use `controller: true` in ownerReferences, enabling Kubernetes garbage collection:

- TEC owns: Deployments, Services, ConfigMaps, Jobs, ApprovedImages
- Machine owns: AttestationKey (transferred from TEC via merge patch once its approved automatically).
- AttestationKey owns: Secret (with controller: true + finalizer)

`generate_owner_reference` renamed to `generate_owner_controller_reference` to reflect that all owner refs are now controller refs with `blockOwnerDeletion: true`.

All controllers are owners, but not all owners have to be controllers. Kubernetes mandates only one controller. Multiple owners are fine for Garbage Collection and delete cascades.

New `apply_resource!` macro replaces `create_or_info_if_exists!` for idempotent resource creation via SSA (PATCH with Patch::Apply).

`update_reference_values`: uses full-data SSA patch on trustee-data ConfigMap (includes all keys: kbs-config.toml, policy.rego, reference-values.json) to prevent the field manager from implicitly relinquishing keys not in a partial patch.

If a field manager (controller owning a field of a CRD)  makes a SSA patch, all fields of that patch are considered owned, and any previously owned fields not part of the patch are released from its control.

`update_attestation_keys`: uses full-object SSA patch on the Deployment (sends entire deployment.spec) to avoid stripping required fields like spec.selector. Metadata fields (managedFields, resourceVersion, uid) are excluded from the SSA body to prevent 400 errors.

`do_mount_secret`: kept as GET+mutate+replace (PUT) since it follows a read-modify-write pattern on the same field manager. Migrating this is significant code rework.

Owner transfer in `approve_ak` uses Merge patch (not SSA) because SSA would upsert into the ownerReferences array, leaving stale references of both TEC. In the end TEC and Machine would co-exist as controllers, which we don't want. Merge patch replaces the entire ownerReferences field atomically.

Single field manager `trusted-cluster-operator` used consistently via `FIELD_MANAGER` constant. register-server uses its own `register-server` field manager for Machine creation (POST with fieldManager query param, not SSA).

`update_status!` macro now requires explicit type and field manager parameters, enabling force-apply on the status subresource where multiple writers exist (e.g. ApprovedImage status).

- Duplicate volume entries: `do_mount_secret` now checks for existing volumes before adding, preventing 422 errors on reconciler retry.
- Secret controller filtering: `update_attestation_keys` and `secret_reconcile` now require `controller: true` on AttestationKey owner refs, not just `kind == "AttestationKey"`.
- ConfigMap data loss: full-data SSA patch on trustee-data prevents kbs-config.toml and policy.rego from being garbage-collected when only reference-values.json is updated.

Added unit tests validating SSA body content, ownership, idempotency, field manager usage, and filtering logic across trustee, reference values, attestation key register, and register server modules.

Commit message co-authored-by: Cursor <cursoragent@cursor.com>

## Summary by Sourcery

Migrate core Kubernetes interactions to server-side apply while standardizing controller ownership and field manager usage across the operator, register-server, and auxiliary binaries.

New Features:
- Introduce an apply_resource! macro to create or update namespaced resources idempotently via server-side apply with a consistent field manager.
- Add a deploy.sh helper script to stand up a kind cluster, build and deploy the operator stack, and create a sample VM for end-to-end validation.

Bug Fixes:
- Prevent duplicate secret volume mounts on trustee pods to avoid reconciliation errors on retries.
- Ensure trustee-data ConfigMap updates preserve existing keys when recomputing reference values, preventing configuration loss.
- Filter attestation key secrets and AK-owned secrets by controller=true owner references to avoid acting on non-controller-owned resources.

Enhancements:
- Adopt controller ownerReferences (controller=true, blockOwnerDeletion=true) across TEC-owned and Machine-owned resources to enable reliable garbage collection and lifecycle management.
- Standardize use of a FIELD_MANAGER constant and SSA-based Patch::Apply for status and spec updates where appropriate, including force-apply where multiple writers exist.
- Use merge patches instead of SSA for AttestationKey owner transfer to Machines to ensure ownerReferences are replaced atomically without leaving stale owners.
- Refine attestation key, trustee, reference value, and register-server tests to validate SSA payloads, ownership semantics, idempotency, and field manager behavior.
- Adjust register-server and attestation-key-register to set fieldManager on creates so the operator can attribute resource ownership correctly.
- Switch trustee deployment volume updates for attestation keys to full-object SSA patches to avoid losing required fields like selectors and to keep base volumes intact.

Tests:
- Extend unit and integration tests around attestation key approval, secret reconciliation, trustee reference value updates, approved image adoption, and machine creation to cover SSA flows, controller ownership, and force-applied status updates.

### DIFF
--- a/attestation-key-register/src/main.rs
+++ b/attestation-key-register/src/main.rs
@@ -16,7 +16,8 @@ use uuid::Uuid;
 
 use trusted_cluster_operator_lib::endpoints::ATTESTATION_KEY_REGISTER_RESOURCE;
 use trusted_cluster_operator_lib::{
-    generate_owner_reference, get_trusted_execution_cluster, AttestationKey, AttestationKeySpec,
+    self, generate_owner_controller_reference, get_trusted_execution_cluster, AttestationKey,
+    AttestationKeySpec,
 };
 
 #[derive(Parser)]
@@ -64,15 +65,17 @@ async fn handle_registration(
 
     let api: Api<AttestationKey> = Api::default_namespaced(client.clone());
 
-    // Get the TrustedExecutionCluster to use as owner reference
+    // TEC is owner-controller of the AttestationKey until the operator approves the key and transfers control to Machine, which then becomes the controller and manages lifecycle of the AttestationKey.
     let cluster = match get_trusted_execution_cluster(client.clone()).await {
         Ok(c) => c,
         Err(e) => return internal_error(e.context("Failed to get TrustedExecutionCluster")),
     };
 
-    let owner_reference = match generate_owner_reference(&cluster) {
+    let owner_controller_reference = match generate_owner_controller_reference(&cluster) {
         Ok(o) => o,
-        Err(e) => return internal_error(e.context("Failed to generate owner reference")),
+        Err(e) => {
+            return internal_error(e.context("Failed to generate owner-controller reference"))
+        }
     };
 
     match api.list(&Default::default()).await {
@@ -104,7 +107,7 @@ async fn handle_registration(
     let attestation_key = AttestationKey {
         metadata: ObjectMeta {
             name: Some(name.clone()),
-            owner_references: Some(vec![owner_reference]),
+            owner_references: Some(vec![owner_controller_reference]),
             ..Default::default()
         },
         spec: AttestationKeySpec {
@@ -114,7 +117,17 @@ async fn handle_registration(
         status: None,
     };
 
-    match api.create(&Default::default(), &attestation_key).await {
+    // Client side apply, as this is a one time operation (no reconciliation loop) for each attestation key, and one owner. SSA is mainly used for patching. Setting field manager, so that we can identify the source of the creation later for reconciliation by operator.
+    match api
+        .create(
+            &kube::api::PostParams {
+                field_manager: Some(trusted_cluster_operator_lib::FIELD_MANAGER.to_string()),
+                ..Default::default()
+            },
+            &attestation_key,
+        )
+        .await
+    {
         Ok(created) => {
             let name = created.metadata.name.unwrap_or_default();
             info!("Successfully created AttestationKey: {name}",);
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -30,13 +30,31 @@ use conditions::*;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::{Condition, OwnerReference, Time};
 use kube::Resource;
 
+pub const FIELD_MANAGER: &str = "trusted-cluster-operator";
+
+// 2 arms for updating status, one for normal updates and one for force updates.
 #[macro_export]
 macro_rules! update_status {
-    ($api:ident, $name:expr, $status:expr) => {{
-        let patch = kube::api::Patch::Merge(serde_json::json!({"status": $status}));
-        $api.patch_status($name, &Default::default(), &patch).await
+    ($api:ident, $name:expr, $status:expr, $type:ty, $field_manager:expr) => {{
+        let patch = kube::api::Patch::Apply(serde_json::json!({
+            "apiVersion": <$type as kube::Resource>::api_version(&()),
+            "kind": <$type as kube::Resource>::kind(&()),
+            "status": $status
+        }));
+        let params = kube::api::PatchParams::apply($field_manager);
+        $api.patch_status($name, &params, &patch).await
             .map_err(Into::<anyhow::Error>::into)
-    }}
+    }};
+    ($api:ident, $name:expr, $status:expr, $type:ty, $field_manager:expr, force) => {{
+        let patch = kube::api::Patch::Apply(serde_json::json!({
+            "apiVersion": <$type as kube::Resource>::api_version(&()),
+            "kind": <$type as kube::Resource>::kind(&()),
+            "status": $status
+        }));
+        let params = kube::api::PatchParams::apply($field_manager).force();
+        $api.patch_status($name, &params, &patch).await
+            .map_err(Into::<anyhow::Error>::into)
+    }};
 }
 
 pub fn condition_status(status: bool) -> String {
@@ -108,8 +126,8 @@ pub fn committed_condition(
     }
 }
 
-/// Generate an OwnerReference for any Kubernetes resource
-pub fn generate_owner_reference<T: Resource<DynamicType = ()>>(
+/// OwnerReference with `controller: true` for lifecycle and garbage collection.
+pub fn generate_owner_controller_reference<T: Resource<DynamicType = ()>>(
     object: &T,
 ) -> anyhow::Result<OwnerReference> {
     let name = object.meta().name.clone();
--- a/operator/src/attestation_key_register.rs
+++ b/operator/src/attestation_key_register.rs
@@ -14,7 +14,7 @@ use k8s_openapi::apimachinery::pkg::{
     util::intstr::IntOrString,
 };
 use kube::{
-    Api, Client, Resource,
+    Api, Client,
     api::{ListParams, ObjectList, Patch, PatchParams},
     runtime::{Controller, controller::Action, finalizer, finalizer::Event, watcher},
 };
@@ -29,7 +29,7 @@ use trusted_cluster_operator_lib::{AttestationKey, AttestationKeyStatus, Machine
 use crate::conditions::attestation_key_approved_condition;
 use crate::trustee;
 use operator::{
-    ControllerError, TLS_DIR, controller_error_policy, create_or_info_if_exists, read_certificate,
+    ControllerError, TLS_DIR, apply_resource, controller_error_policy, read_certificate,
     upsert_condition,
 };
 
@@ -97,7 +97,7 @@ pub async fn create_attestation_key_register_deployment(
         ..Default::default()
     };
 
-    create_or_info_if_exists!(client, Deployment, deployment);
+    apply_resource!(client, Deployment, deployment);
     info!("Attestation key register deployment created successfully");
     Ok(())
 }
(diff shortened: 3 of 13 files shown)


SpaceFace02:
Unit tests passing:
image

Integration tests passing:
Screenshot From 2026-06-02 13-22-26

lib/src/lib.rs
```diff
@@ -30,13 +30,31 @@ use conditions::*;
...
+        let patch = kube::api::Patch::Apply(serde_json::json!({
+            "apiVersion": <$type as kube::Resource>::api_version(&()),
+            "kind": <$type as kube::Resource>::kind(&()),
+            "status": $status
+        }));
+        let params = kube::api::PatchParams::apply($field_manager);
+        $api.patch_status($name, &params, &patch).await
+            .map_err(Into::<anyhow::Error>::into)
+    }};
+    ($api:ident, $name:expr, $status:expr, $type:ty, $field_manager:expr, force) => {{
```
Jakob-Naucke, line 48: So this sets the component (operator, compute-pcrs) as field manager for the status, but also just uses `force` the only time that it could actually conflict. I don't really see the downside of having just one field manager by that point but I'm open to hearing @alicefr's opinion before you refactor.

lib/src/lib.rs
```diff
@@ -108,8 +126,8 @@ pub fn committed_condition(
     }
 }
 
-/// Generate an OwnerReference for any Kubernetes resource
-pub fn generate_owner_reference<T: Resource<DynamicType = ()>>(
+/// OwnerReference with `controller: true` for lifecycle and garbage collection.
+pub fn generate_owner_controller_reference<T: Resource<DynamicType = ()>>(
```
Jakob-Naucke, line 130: not sure why this is necessary, it's not like there's a new function that this function's name must disambiguate from

attestation-key-register/src/main.rs
```diff
@@ -64,15 +65,17 @@ async fn handle_registration(
 
     let api: Api<AttestationKey> = Api::default_namespaced(client.clone());
 
-    // Get the TrustedExecutionCluster to use as owner reference
+    // TEC is owner-controller of the AttestationKey until the operator approves the key and transfers control to Machine, which then becomes the controller and manages lifecycle of the AttestationKey.
```
Jakob-Naucke, line 68: What's an owner-controller?

operator/src/attestation_key_register.rs
```diff
@@ -405,4 +420,243 @@ mod tests {
...
+            },
+            spec: trusted_cluster_operator_lib::MachineSpec {
+                id: "machine-uuid".to_string(),
+            },
+            status: None,
+        }
+    }
+
+    #[tokio::test]
+    async fn test_approve_ak_full_flow() {
```
Jakob-Naucke, line 455: It's generally good to expand the tests here. Maybe put new tests in a separate commit because this commit's almost unreadably big

attestation-key-register/src/main.rs
SpaceFace02: Previously, we had the function ```generate_owner_reference``` which set controller to true.

This is ambiguous, because not all owners are controllers. A k8s object can have only 1 controller in its set of owner references ([here](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#OwnerReference)). It can however have multiple non-controller linking owner references.

operator/src/attestation_key_register.rs
SpaceFace02: Sure, had kept unit tests in the same commit as they test the same functionality introduced and made sense to be together. But I will split them for readability and ease of review.

lib/src/lib.rs
SpaceFace02: Previously, we had multiple writers to the Status Field of the ApprovedImage. the ComputePCR also updated the status with committed_condition (while status as PCR's are computing), and the operator requeued every 10 seconds to check whether the computation had completed and then updated the same field.

To fix this, ideally we should have only the operator update the status of the field. We can discuss this design decision in detail with Alice.

SpaceFace02: >>  I don't really see the downside of having just one field manager by that point

Field manager is inherently a SSA concept, initially we had 2 writers to the same field, and updated it client side. There was no concept of field managers, because the client patch would overwrite any changes, even if written previously by another writer.

lib/src/lib.rs
SpaceFace02: Mainly to disambiguate between a controller owner, and a linking owner.

operator/src/trustee.rs
```diff
@@ -163,10 +173,17 @@ pub async fn do_mount_secret(client: Client, id: &str, add: bool) -> Result<()>
...
+    // Additional check to prevent duplicate mounts.
     if add {
-        let (volume, volume_mount) = generate_secret_volume(id);
-        pod_spec.volumes.get_or_insert_default().push(volume);
-        vol_mounts.push(volume_mount);
+        let already_mounted = pod_spec
+            .volumes
+            .as_ref()
+            .is_some_and(|vs| vs.iter().any(|v| v.name == id));
+        if !already_mounted {
```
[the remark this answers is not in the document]
SpaceFace02: @Jakob-Naucke  Thoughts on this issue?

alicefr:
@SpaceFace02 please sign the commit. We should add a check that verifies that all the commits are signed. Filed https://github.com/trusted-execution-clusters/operator/issues/270

alicefr:
@SpaceFace02 is there a way to split the single commit into multiple. This is kind of hard to review it

SpaceFace02:
> @SpaceFace02 is there a way to split the single commit into multiple. This is kind of hard to review it

Have split the commits into 3:
One for feature changes
one for uts
one for integration tests

operator/src/trustee.rs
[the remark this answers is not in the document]
SpaceFace02: Volumes and volume mounts update in one transaction/patch atomically. Partial updates won't be supported.

attestation-key-register/src/main.rs
Jakob-Naucke: if we keep these renames and comment changes, I think they should be a separate commit. not sure if we do though, cc @alicefr

operator/src/attestation_key_register.rs
```diff
@@ -419,4 +420,243 @@ mod tests {
...
+        // approve_ak with no prior status, no Machine owner, and secret not existing:
+        // 1. PATCH /status (SSA status update). AttestationKey is approved.
+        // 2. PATCH owner transfer (Merge). Ownership transfered to Machine.
+        // 3. GET secret (check existence)
+        // 4. PATCH secret (apply_resource! SSA create)
```
Jakob-Naucke, line 460: the numbers and HTTP endpoints are in the code, so I think this is just drift risk. if you like you can add the comment as to what happens at those respective points.

operator/src/reference_values.rs
```diff
@@ -557,4 +586,112 @@ mod tests {
             assert!(disallow_image(ctx, "registry").await.is_ok());
         });
     }
+
+    // SSA related unit test helpers.
+
```
Jakob-Naucke, line 591: can drop, it's just one and could be reused

Jakob-Naucke (commented):
Thank you for the separate commits! However, when a test update is necessary to how the code changes, it should be one commit so that bisects stay intact.

operator/src/trustee.rs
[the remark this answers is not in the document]
Jakob-Naucke: yea I don't think it's necessary, will also change with #248

attestation-key-register/src/main.rs
alicefr: yes, please try to generate clean commits it gets easier to review

lib/src/lib.rs
alicefr: > To fix this, ideally we should have only the operator update the status of the field. We can discuss this design decision in detail with Alice.

Yeah, we need to revisit the design of the approved images ideally, there should be a single writer. Let's discuss it today

PR #273, opened 2026-06-04
Register AK via KBS API

iroykaufman:
Following [PR#248](https://github.com/trusted-execution-clusters/operator/pull/248)
This PR introduces the registration of AK via the KBS API.

- Add a `register_ak` function to add the AK to the KBS as a reference value.
- Update the attestation policy with a check to verify that the AK is registered.
- Add a test to confirm the creation and synchronization of the AK upon trustee restarts.

## Summary by Sourcery

Register trustee machine secrets and attestation keys with the KBS API and keep them synchronized across trustee restarts.

New Features:
- Send machine LUKS secrets to the KBS via its API instead of mounting them into the trustee pod.
- Register attestation keys in the KBS as trusted reference values and enforce their presence in the attestation policy.
- Introduce JWT-based admin authentication and authorization for the KBS using an operator-managed Ed25519 key pair.
- Add a controller to resync machine LUKS secrets and attestation keys to the KBS when the trustee deployment becomes ready.

Enhancements:
- Upgrade the trustee (KBS) image and reconfigure it to use KV storage, authenticated admin API, and updated attestation token handling.
- Generate and manage trustee auth keys as a Kubernetes Secret and mount the public key into the trustee pod for KBS admin auth.
- Ensure TLS/crypto providers are explicitly installed to avoid conflicts and make tests compatible with the new crypto stack.

Build:
- Pin compute-pcrs-lib to a specific revision for reproducible builds.
- Update the default trustee image tag used in Makefile to a newer KBS version.

Tests:
- Add integration tests verifying that machine LUKS secrets are synced to the KBS, resynced after trustee restarts, and cleaned up after machine deletion.
- Add integration tests verifying that attestation keys are registered with the KBS and re-registered after trustee restarts.
- Add unit tests for Ed25519 key pair generation utilities.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,71 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "aead"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
+dependencies = [
+ "crypto-common 0.1.7",
+ "generic-array",
+]
+
+[[package]]
+name = "aes"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


iroykaufman:
[PR#248](https://github.com/trusted-execution-clusters/operator/pull/248) include this PR change

PR #266, opened 2026-05-29
tests: Create passthrough route for OpenShift

Jakob-Naucke:
expose doesn't work with certificates. drive-by: use oc for OpenShift backend

## Summary by Sourcery

Update Kubernetes test utilities to support platform-specific CLI commands and add passthrough route exposure on OpenShift using service ports.

New Features:
- Create passthrough routes for OpenShift services using the configured service port instead of relying on generic expose functionality.

Enhancements:
- Route kubectl/oc invocations in test utilities through a K8sPlatform abstraction, allowing OpenShift to use oc while other platforms use kubectl.
- Use named constants for deployment and service identifiers and ports when waiting for deployments and exposing services in tests.

PR #210, opened 2026-03-06
tests: Prepare upstream Azure tests

Jakob-Naucke:
- track some tools' versions
- optimize Makefile for partial & parallel installations
- enable Kind forwarding in tests

See https://github.com/openshift/release/pull/75842 for how I plan to use it

PR #285, opened 2026-06-19
Gear for OpenShift CI

Jakob-Naucke:
- Some reversals of #210 as at least the Azure CI will, after all, not rely on Kind
- Use `await_condition` instead of polling when possible. Is also more flexible for deployments that don't exist yet.
- Should use loadbalancing services, not passthrough like in #266
- #246 introduced some finalizer usage that requires RBAC on OpenShift
- Use user-data for Azure, custom data is sometimes not yet available at fetch stage
- Wait for custom resources removal, do not remove explicitly like in #278
- small fixes
---
- See https://github.com/openshift/release/pull/79393 (Azure) & https://github.com/openshift/release/pull/80214 (KubeVirt) for testing progress
- Saw some issues believed to be related to Trustee pod restarts on the KubeVirt rehearsals. Should merge #248 before continuing there, but not a blocker for this PR.
- Review commit-by-commit is advised

## Summary by Sourcery

Adapt the operator, tests, and tooling for running on OpenShift-based CI while modernizing Kubernetes waiting logic and simplifying dependencies.

New Features:
- Expose operator services on OpenShift via LoadBalancer services with certificate-driven endpoints instead of Routes.
- Update TrustedExecutionCluster instances dynamically with the externally reachable trustee address once services are exposed.

Bug Fixes:
- Ensure Azure VMs in tests use cloud-init user-data instead of custom-data so configuration is reliably available.
- Add RBAC permissions for all custom resource finalizers to avoid deletion failures on OpenShift clusters.
- Fix default component image tag formatting by including the leading 'v' to match published images.

Enhancements:
- Replace custom polling helpers in tests with kube-runtime await_condition and timeouts for more robust, race-free readiness checks.
- Wait for CRD-backed resources to be garbage-collected rather than deleting them manually in tests, reducing flakiness.
- Refactor operator component installation to fail fast with contextual errors and trigger reconcile requeues on partial installation.
- Derive OpenShift service endpoints from LoadBalancer status instead of OpenShift Route and Ingress APIs, removing those dependencies.
- Simplify Makefile tooling and Go module dependencies by dropping Kind and OpenShift-specific APIs and improving kopium version discovery.

Tests:
- Revise integration tests to use condition-based waiting for deployments, custom resources, and ConfigMaps, improving stability under CI load.
- Adjust test cleanup logic to verify absence of Machines, ApprovedImages, AttestationKeys, and related Secrets instead of force-deleting them.


operator/src/main.rs
```diff
@@ -35,7 +35,7 @@ use operator::*;
 /// Default tag for Trustee image
 const TRUSTEE_VERSION: &str = "v0.17.0";
 /// Default version tag for operator-managed component images
-const COMPONENT_VERSION: &str = "0.2.0";
+const COMPONENT_VERSION: &str = "v0.2.0";
```
alicefr, line 38: Not related to this PR, but the versioning is actually wrong since we already have the release 0.2.1. This should be latest and we should set dynmically when we do the release. Filed https://github.com/trusted-execution-clusters/operator/issues/286

test_utils/src/lib.rs
```diff
@@ -267,6 +274,27 @@ impl K8sPlatform for Kind {
...
+    None,
+}
+
+impl OpenShift {
+    async fn get_url(&self, client: &Client, namespace: &str, service: &str) -> OpenShiftHost {
+        let services: Api<Service> = Api::namespaced(client.clone(), namespace);
+        let Ok(svc) = services.get(service).await else {
+            return OpenShiftHost::None;
+        };
+        let ingress = &svc.status.unwrap().load_balancer.unwrap().ingress.unwrap()[0];
```
alicefr, line 282: Are those unwraps safe? All those fields always set?

test_utils/src/lib.rs
```diff
@@ -278,22 +306,67 @@ impl K8sPlatform for OpenShift {
 
     async fn expose(
         &self,
-        _: &Client,
+        client: &Client,
         namespace: &str,
         service: &str,
+        deployment: &str,
+        cert_name: &str,
         _: &str,
```
alicefr, line 305: can we simplify the signature  of this function? parameter seems a lot to me

test_utils/src/lib.rs
```diff
@@ -932,68 +959,63 @@ impl TestContext {
...
-            (ATTESTATION_KEY_REGISTER_SERVICE, ak_port),
-            (REGISTER_SERVER_SERVICE, REGISTER_SERVER_PORT),
         ] {
-            platform
-                .expose(&self.client, ns, svc, &self.test_name, port)
-                .await?;
+            let info = format!("Waiting for deployment {depl} to be ready");
+            test_info!(&self.test_name, "{info}");
+            let done = await_condition(depls.clone(), depl, depl_ready);
+            timeout(scaled_duration(300), done).await??;
```
alicefr: nit: 
```suggestion
timeout(scaled_duration(300), done).await.context(ctx)??;
```

operator/src/main.rs
```diff
@@ -132,9 +132,14 @@ async fn reconcile(
...
     }
 
-    install_trustee_configuration(kube_client.clone(), &cluster).await?;
-    install_register_server(kube_client.clone(), &cluster).await?;
-    install_attestation_key_register(kube_client.clone(), &cluster).await?;
+    if let Err(e) = install_components(&kube_client, &cluster).await {
+        // warn with `:?` to also get context
+        warn!(
+            "Installation of a component failed: {e:?}\n \
+             This can occur due to owner references to incompletely propagated CRDs. Requeueing..."
```
alicefr: what if it is another type of error? You are returning an error also on creation for example

test_utils/src/lib.rs
Jakob-Naucke: Don't know for sure but I do think it's fatal to the test

operator/src/main.rs
Jakob-Naucke: This is intended for on creation. Do you mean that we should have a maximum failure and installation failed status?

alicefr: No, just if the only cause for the error can be the owner reference or maybe something like create the one of  the object because it already exists

Jakob-Naucke: for "if exists", we have `create_or_info_if_exists`, but there are more possible errors here. ideally, the second line of the warning would only be present in the specific case I saw this, but then you're getting into error parsing.

alicefr: Maybe just keep it generic and just print the error

test_utils/src/lib.rs
```diff
@@ -276,76 +299,100 @@ impl K8sPlatform for OpenShift {
...
+        services.patch(service, &pp, &Patch::Merge(&json)).await?;
+        let has_ingress = |svc: Option<&Service>| {
+            let chk_lb = |bal: &LoadBalancerStatus| bal.ingress.is_some();
+            let chk_st = |st: &ServiceStatus| st.load_balancer.as_ref().map(chk_lb);
+            let chk_svc = |svc: &Service| svc.status.as_ref().and_then(chk_st);
+            svc.and_then(chk_svc).unwrap_or(false)
+        };
+        let ingress_ready = await_condition(services, service, has_ingress);
+        let ctx = format!("waiting for ingress on {service} to be ready");
+        let duration = scaled_duration(60);
```
SpaceFace02, line 323: How exactly are we estimating the multiplier scale env and scale timeout duration across all files? Is it a ballpark estimate or are we using some heuristic?

Jakob-Naucke: These are not backed by measurements or anything; they're set based on experience and can be increased (if we see a failure being an issue of merely a timeout that's too short) or decreased (if a test frequently takes more time to fail than it would need to).

Similar story with the multipliers, just a tradeoff of how fast is your environment/how long are you willing to wait

alicefr (approved):

PR #302, opened 2026-07-01
rvs: Fix race of ApprovedImage adoption with TEC object deletion

iroykaufman:
The race scenario:
- Assume the operator is installed and TEC object exists
- Apply an approved-image yaml file (create a new CR instance)
- Delete the TEC object (immediately)
- The new approved-image is not deleted with the TEC object

This happens because kube-rs finalizer() does not call the Apply handler on the first reconcile — it only adds the finalizer and returns. This means adopt_approved_image inside image_add_reconcile never runs on the first reconcile. If the TEC is deleted before the second reconcile, the owner reference is never set and GC (garbage collector) cannot cascade-delete the ApprovedImage.

Move adoption before the finalizer() call in image_reconcile so it runs on the very first reconcile.

For more information see https://docs.rs/kube/latest/kube/runtime/finalizer/fn.finalizer.html in sections Guarantees, Assumptions (Third point) and Expected Flow(from 1 to 6).

## Summary by Sourcery

Ensure ApprovedImage resources are adopted by the owning TrustedExecutionCluster on first reconcile to avoid orphaned images during TEC deletion, and remove the now-unnecessary bulk adoption logic from the main reconcile path.

Bug Fixes:
- Fix race where ApprovedImage resources could miss owner references and not be garbage-collected when their TrustedExecutionCluster is deleted.

Enhancements:
- Adopt ApprovedImage resources within the image reconcile flow before finalizer handling, instead of via a separate bulk adoption step in the main reconcile.
- Simplify tests and controller interactions by removing bulk ApprovedImage adoption and its dedicated test coverage.

### DIFF
--- a/operator/src/reference_values.rs
+++ b/operator/src/reference_values.rs
@@ -247,6 +247,23 @@ async fn image_reconcile(
         .await
         .map_err(|e| -> ControllerError { e.into() })?;
 
+    let uid_owns = |uid: &String| {
+        let refs = image.metadata.owner_references.as_ref();
+        refs.map(|os| os.iter().any(|o| o.uid == *uid))
+    };
+    let cluster_owns = |cluster: &TrustedExecutionCluster| {
+        let uid = cluster.metadata.uid.as_ref();
+        uid.and_then(uid_owns).unwrap_or(false)
+    };
+
+    if let Some(ref cluster) = cluster
+        && !cluster_owns(cluster)
+    {
+        adopt_approved_image(kube_client.clone(), &name, cluster)
+            .await
+            .map_err(|e| -> ControllerError { e.into() })?;
+    }
+
     let images: Api<ApprovedImage> = Api::default_namespaced(kube_client.clone());
     finalizer(&images, APPROVED_IMAGE_FINALIZER, image, |ev| async {
         match ev {
@@ -277,19 +294,6 @@ async fn image_add_reconcile(
         info!("TrustedExecutionCluster is being deleted, deferring image processing for {name}");
         return Ok(Action::requeue(Duration::from_secs(5)));
     }
-    let uid_owns = |uid: &String| {
-        let refs = image.metadata.owner_references.as_ref();
-        refs.map(|os| os.iter().any(|o| o.uid == *uid))
-    };
-    let cluster_owns = |cluster: &TrustedExecutionCluster| {
-        let uid = cluster.metadata.uid.as_ref();
-        uid.and_then(uid_owns).unwrap_or(false)
-    };
-    // Adopt the image by adding TEC as owner reference if not already owned
-    if !cluster_owns(&cluster) {
-        adopt_approved_image(client.clone(), name, &cluster).await?;
-    }
-
     let (action, reason) = match handle_new_image(client.clone(), image).await {
         Ok(reason) => (Action::await_change(), reason),
         Err(e) => {


operator/src/reference_values.rs
Jakob-Naucke, line 246: I know I approved this in #273 and I'm sorry that I'm only complaining now, but I did wonder about something now that we have space to discuss it, so thank you for raising separately. Thus:

Is this really an improvement? When image creation and cluster deletion fall close to one another, this `cluster` might also be `None`, to the same effect.

Or did this cause a test failure? If it did, I wonder if that test should lower its expectations on how quick an adoption happens. `test_approved_image_readoption` has been doing that since https://github.com/trusted-execution-clusters/operator/commit/66640711c998e72bfc16271715ab6ccd04f1b9a0 (which was only merged a few days ago -- did this patch make a test pass more consistently since that?)

iroykaufman: If the cluster has been deleted before adopt image was called, then yes, we will have the same problem. But we can add a condition where, if the cluster has been deleted and the image has not adopted yet, we will delete the image.
Before this patch, the test failed on `test_approved_image_readoption`. I don't remember if I used https://github.com/trusted-execution-clusters/operator/commit/66640711c998e72bfc16271715ab6ccd04f1b9a0 but waiting for a bit longer probably can solve it but in my opinion it's not an ideal solution. WDYT

Jakob-Naucke: > If the cluster has been deleted before adopt image was called, then yes, we will have the same problem. But we can add a condition where, if the cluster has been deleted and the image has not adopted yet, we will delete the image. Before this patch, the test failed on `test_approved_image_readoption`.

hmm the intention is that you may have ApprovedImages before a TrustedExecutionCluster exists, but with your suggestion (IIUIC) they'd always be removed, even if a TrustedExecutionCluster had never existed

> I don't remember if I used [6664071](https://github.com/trusted-execution-clusters/operator/commit/66640711c998e72bfc16271715ab6ccd04f1b9a0) but waiting for a bit longer probably can solve it but in my opinion it's not an ideal solution. WDYT

FWIW that commit doesn't wait a fixed time in seconds, but awaits the ownership

operator/src/reference_values.rs
```diff
@@ -247,6 +231,23 @@ async fn image_reconcile(
...
+    };
+    let cluster_owns = |cluster: &TrustedExecutionCluster| {
+        let uid = cluster.metadata.uid.as_ref();
+        uid.and_then(uid_owns).unwrap_or(false)
+    };
+
+    if let Some(ref cluster) = cluster
+        && !cluster_owns(cluster)
+    {
+        adopt_approved_image(kube_client.clone(), &name, cluster)
```
alicefr, line 262: We are patching twice the Approve image. I'm wondering if we could refactor the adopt_approved_image and remove the patch from there and do the modification together with the finalizer in the main reconcile loop. It should save a reconciliation loop plus a write operation.

alicefr:
Thanks for raising the PR, I like the fact that we are now adopting the image as part of its reconciliation loop. @Jakob-Naucke I know I introduce this, but probably we have the same pattern also on other object we should check it.

Jakob-Naucke:
> I like the fact that we are now adopting the image as part of its reconciliation loop

I realize the tests passed on this one, but logically, should the image even be adopted as part of its reconciliation when a cluster is created? What if the cluster is created later?

alicefr:
> > I like the fact that we are now adopting the image as part of its reconciliation loop
> 
> I realize the tests passed on this one, but logically, should the image even be adopted as part of its reconciliation when a cluster is created? What if the cluster is created later?

Ah I see, you mean it should be part of the reconciliation loop of TEC right?  I guess it should be on both depending who is created first

Jakob-Naucke:
> > I realize the tests passed on this one, but logically, should the image even be adopted as part of its reconciliation when a cluster is created? What if the cluster is created later?
> 
> Ah I see, you mean it should be part of the reconciliation loop of TEC right? I guess it should be on both depending who is created first

Yes, and I didn't clarify this in my earlier answer, but this is already what we do today: https://github.com/trusted-execution-clusters/operator/blob/103fa3f9d10a697cf3ad2ad40543cc94fffce217/operator/src/main.rs#L140 https://github.com/trusted-execution-clusters/operator/blob/103fa3f9d10a697cf3ad2ad40543cc94fffce217/operator/src/reference_values.rs#L290

alicefr:
> Yes, and I didn't clarify this in my earlier answer, but this is already what we do today:
> 
> https://github.com/trusted-execution-clusters/operator/blob/103fa3f9d10a697cf3ad2ad40543cc94fffce217/operator/src/main.rs#L140
> 
> https://github.com/trusted-execution-clusters/operator/blob/103fa3f9d10a697cf3ad2ad40543cc94fffce217/operator/src/reference_values.rs#L290

We probably need both. Roy's fix looks correct to me but we need to leave the adoption of the approved images also in the TEC reconcile loop

iroykaufman:
> > Yes, and I didn't clarify this in my earlier answer, but this is already what we do today:
> > https://github.com/trusted-execution-clusters/operator/blob/103fa3f9d10a697cf3ad2ad40543cc94fffce217/operator/src/main.rs#L140
> > 
> > https://github.com/trusted-execution-clusters/operator/blob/103fa3f9d10a697cf3ad2ad40543cc94fffce217/operator/src/reference_values.rs#L290
> 
> We probably need both. Roy's fix looks correct to me but we need to leave the adoption of the approved images also in the TEC reconcile loop

This make sense. I'm on it.

Jakob-Naucke (approved):
Fine. I still think https://github.com/trusted-execution-clusters/operator/commit/66640711c998e72bfc16271715ab6ccd04f1b9a0 fixed the test failure that this patch tries to improve, and that the race is still possible after this patch, but can see how it might be less likely.

operator/src/reference_values.rs
alicefr: Nevermind, we need to reimplement finalizer otherwise.

alicefr (approved):

Jakob-Naucke:
there's one failing test but we removed it in the meantime, we could rebase but IMO we can just merge @alicefr

Jakob-Naucke:
urgh, ironically, an ApprovedImage stayed behind on a test here…

```
INFO: test_vm_reboot_attestation: TrustedExecutionCluster trusted-execution-cluster has been deleted
Error: Resources were left behind after 60s, last error was: Resource still present: ApprovedImage
```

I have some experimental maybe-fixes on https://github.com/Jakob-Naucke/trusted-cluster-operator/tree/shutdown-timeout but ideally I should spend more time on thinking what, if any of these and/or this PR, is the right way to fix this

iroykaufman:
> urgh, ironically, an ApprovedImage stayed behind on a test here…
> 
> ```
> INFO: test_vm_reboot_attestation: TrustedExecutionCluster trusted-execution-cluster has been deleted
> Error: Resources were left behind after 60s, last error was: Resource still present: ApprovedImage
> ```
> 
> I have some experimental maybe-fixes on https://github.com/Jakob-Naucke/trusted-cluster-operator/tree/shutdown-timeout but ideally I should spend more time on thinking what, if any of these or this PR, is the right way to fix this

Yes, this is very unstable. This is the second time I ran the test, and last time it failed on a different test with a different error. As we discussed [here](https://github.com/trusted-execution-clusters/operator/pull/302#discussion_r3504212246), the race still exists, though we expected it to be less likely. I'll try spending more time investigating this.

PR #303, opened 2026-07-01
gha: Remove `pull_request_target` flow

Jakob-Naucke:
was added in error, thinking that the `labeled` event could not apply to `pull_request` when it can. This lead to failures because checkout is not enabled by default on `pull_request_target` for security reasons, as happened in #302.

## Summary by Sourcery

CI:
- Update the integration tests GitHub Actions workflow to trigger on labeled pull_request events alongside the standard pull_request event types.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -5,10 +5,8 @@
 name: "Integration tests"
 on:
   pull_request:
-    branches:
-      - "main"
-  pull_request_target:
-    types: [labeled]
+    # Default types, plus `labeled` for when `ok-to-test` was added
+    types: [opened, synchronize, reopened, labeled]
     branches:
       - "main"
 permissions:


.github/workflows/integration-tests.yml
```diff
@@ -5,10 +5,8 @@
 name: "Integration tests"
 on:
   pull_request:
-    branches:
-      - "main"
-  pull_request_target:
-    types: [labeled]
+    # Default types, plus `labeled` for when `ok-to-test` was added
+    types: [opened, synchronize, reopened, labeled]
```
alicefr, line 9: I know it was already there, but then does it run for every labels right now not only ok-to-test

Jakob-Naucke: no, that we gate like this: https://github.com/trusted-execution-clusters/operator/blob/103fa3f9d10a697cf3ad2ad40543cc94fffce217/.github/workflows/integration-tests.yml#L31-L35

alicefr: ok it's there

alicefr (approved):

Jakob-Naucke:
took a no-op push because old workflow is still active from the label

PR #305, opened 2026-07-02
Added commit signature verification with verified commit

SpaceFace02:
Have tested with unverified commit

## Summary by Sourcery

Enforce commit signature and authorship verification in all CI workflows before running tests, builds, and linting.

CI:
- Add a reusable commit verification workflow that checks PR commits are GPG/SMIME verified and authored by the PR owner.
- Wire the new commit verification job into Rust, Go, integration test, and lint GitHub Actions workflows as a prerequisite step.

### DIFF
--- a/.github/workflows/commit-verification.yml
+++ b/.github/workflows/commit-verification.yml
@@ -1,0 +1,36 @@
+# SPDX-FileCopyrightText: Chirag Rao <crao@redhat.com>
+#
+# SPDX-License-Identifier: CC0-1.0
+
+name: "Commit Verification"
+on:
+  pull_request:
+    branches:
+      - "main"
+  workflow_call:
+permissions:
+  contents: "read"
+
+jobs:
+  verify-commits:
+    name: "Verify commit signatures and authorship"
+    runs-on: " ubuntu-slim"
+    steps:
+      - name: "Verify all PR commits are signed and authored by PR owner"
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_AUTHOR: ${{ github.event.pull_request.user.login }}
+        run: |
+          set -euo pipefail
+          errors=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/commits" \
+             -q '
+              .[] | .sha[:12] as $s |
+              (if .commit.verification.verified != true
+               then "::error::Commit \($s) is not signed and/or not verified (\(.commit.verification.reason))"
+               else empty end)
+            ')
+          if [ -n "${errors}" ]; then
+            echo "${errors}"
+            exit 1
+          fi
+          echo "All commits are signed and verified."
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -16,7 +16,10 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  verify-commits:
+    uses: ./.github/workflows/commit-verification.yml
   build:
+    needs: verify-commits
     name: "Build"
     runs-on: "ubuntu-24.04"
     steps:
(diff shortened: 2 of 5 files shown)


.github/workflows/commit-verification.yml
```diff
@@ -0,0 +1,39 @@
...
+    branches:
+      - "main"
+  workflow_call:
+permissions:
+  contents: "read"
+
+jobs:
+  verify-commits:
+    name: "Verify commit signatures and authorship"
+    runs-on: "ubuntu-24.04"
```
Jakob-Naucke: nit: test should start running a little faster with ubuntu-slim (container-based)

.github/workflows/commit-verification.yml
```diff
@@ -0,0 +1,39 @@
...
+          PR_AUTHOR: ${{ github.event.pull_request.user.login }}
+        run: |
+          set -euo pipefail
+          errors=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/commits" \
+             -q '
+              .[] | .sha[:12] as $s |
+              (if .commit.verification.verified != true
+               then "::error::Commit \($s) is not signed (\(.commit.verification.reason))"
+               else empty end),
+              (if (.author.login // "") != env.PR_AUTHOR
```
Jakob-Naucke: So you can't send a series where you picked someone else's commit? I think that's an important use case to keep

.github/workflows/go.yml
Jakob-Naucke, line 1: I gather you want the verification check to run before other tests, but is there a more unified way to order tests? (e.g. can we design this in a way that we won't forget the same gate when we add another test, or that we don't again need to edit 4 files if we want another test that goes before unit tests)

Then again, as long as GHA remains free for public repos, there isn't even a real incentive to do this itfp

.github/workflows/commit-verification.yml
SpaceFace02: hmm fair, it could cause problems during rebasing or backporting

.github/workflows/go.yml
SpaceFace02: Yeah, might not be needed because we aren't the ones paying for the GHA minutes anyways, but generally a good idea to try to preserve compute if we can as a community courtesy.

As these workflows are seperate files, GH doesn't yet have a way for like a global pre-check. There seems to be a way with Branch Protection Rules, although I can't comment on that

.github/workflows/commit-verification.yml
```diff
@@ -0,0 +1,39 @@
...
+               then "::error::Commit \($s) author (\(.author.login // "unknown")) does not match PR owner (\(env.PR_AUTHOR))"
+               else empty end)
+            ')
```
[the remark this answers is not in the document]
SpaceFace02: pipefail should cover this case I believe, its a false positive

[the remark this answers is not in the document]
Jakob-Naucke: ah I see, although it's `set -e` that covers this, not pipefail

.github/workflows/go.yml
Jakob-Naucke: I'm kind of against this, not just because it makes the files fatter, but also because I saw now it runs per flow, so you get

- Commit Verification / Verify commit signatures and authorship
- Go / Build (pull_request)
- Go / verify-commits / Verify commit signatures and authorship (pull_request)
- Lint / Lints, pinned toolchain (pull_request)
- Lint / verify-commits / Verify commit signatures and authorship (pull_request)

etc.

e: so the "less compute time" argument is negated a bit anyhow

SpaceFace02: Yeah, it does run for each flow, although its relatively lightweight than running the jobs without gating, I think removing it is fine for now.

I am also open to closing this issue as the repo owner can configure both options(verify signatures and prevent merging until pipeline succeeds) from the Branch Protection Rules, although it will run on merge, not on a PR.

SpaceFace02: @alicefr What do you think?

Jakob-Naucke:
Implemented by branch protection rules

PR #308, opened 2026-07-10
Added cargo test for checking compliance to fedora-crypto-policies

### LINKED ISSUE #301: Extend no-rustls test
We test that rustls isn't used: https://github.com/trusted-execution-clusters/operator/blob/f652cc0cb40e5bc9731cd22b09f691227b603a66/.github/workflows/lint.yml#L58 but for easier downstreaming, we want to make sure we're align with https://gitlab.com/redhat-crypto/fedora-crypto-policies at all times.

The Fedora Crypto Policies are an allowlist of backends. However, only a blocklist is easy to implement:
- One approach is to add popular crates to the blocklist based on https://crates.io/keywords/crypto and https://crates.io/categories/cryptography. One crate that we identified in #248 with unclear Fedora alignment is aws-lc-rs.
- A more sophisticated approach would be checking if a crate falls under the categories of the bullet above (keyword "crypto" or category "cryptography") and blocking all those except allowed ones (like openssl)

If more logic is needed, the test could be expanded into a test written in Rust.

SpaceFace02:
Added a test to check for compliance to fedora-crypt-policies for the final shipped bundle, without test-suite deps and build deps.

the allowlist is subject to change based on inputs.

## Summary by Sourcery

Add an automated check ensuring shipped binaries do not depend on non-compliant cryptography crates.

Tests:
- Add a cargo test that scans normal (non-dev, non-build) dependencies for disallowed crypto/TLS crates based on Fedora crypto policies.
- Wire the new crypto-compliance test into the lint GitHub Actions workflow to run on CI.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -55,5 +55,5 @@ jobs:
         run: go vet ./...
       - name: "Ensure Rust & Go conditions definitions use same strings"
         run: cargo test --test equal_conditions
-      - name: "Ensure rustls isn't used"
-        run: "! { cargo tree -e normal | grep rustls | grep -v rustls-pki-types; }"
+      - name: "Ensure no disallowed crypto crates are shipped"
+        run: cargo test --test no_disallowed_crypto
--- a/tests/no_disallowed_crypto.rs
+++ b/tests/no_disallowed_crypto.rs
@@ -1,0 +1,180 @@
+// SPDX-FileCopyrightText: Chirag Rao <crao@redhat.com>
+//
+// SPDX-License-Identifier: MIT
+
+//! Fedora Crypto Policies (<https://gitlab.com/redhat-crypto/fedora-crypto-policies>) only curates policy for an allowlist of backends (OpenSSL, GnuTLS, NSS, libkrb5, ...)
+//! Any crate that reimplements its own crypto/TLS bypasses the system-wide policy and should be blocked.
+
+//! What this test does:
+//! Uses `cargo tree -e normal` to find every crate that actually ships in the built binaries (excluding dev-/build-only dependencies)
+//! `cargo metadata` to look up each one's crates.io categories/keywords. Each crypto crate available in Cargo ecosystem is blocked by default, unless allowlisted.
+//! This test can run air-gapped as all information about the Cargo crates is available in the metadata and the .lock files.
+
+use serde_json::Value;
+use std::collections::HashSet;
+use std::process::Command;
+
+// Crypto crates allowlisted by Fedora Crypto Policies having approved policies and backends.
+/// Crates having only types/errors rather than a crypto/TLS implementation of their own are also allowed (rustls-pki-types).
+const ALLOWED_CRYPTO_CRATES: &[&str] = &[
+    "openssl",
+    "openssl-sys",
+    "rustls-pki-types",
+    "sequoia-openpgp",
+    // Dependencies of openssl, no harm in allowlisting these.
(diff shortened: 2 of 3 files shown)


tests/no_disallowed_crypto.rs
```diff
@@ -0,0 +1,178 @@
...
+// Exclude test-only workspace members which are never shipped to the bundle.
+const TEST_ONLY_WORKSPACE_MEMBERS: &[&str] = &[
+    "trusted-cluster-operator-tests",
+    "trusted-cluster-operator-test-utils",
+];
+
+// Names of every crate reachable via a *normal* (non-dev, non-build) dependency edge from any workspace member that actually ships - i.e. what actually ends up in the built binaries.
+fn normal_dependency_names() -> HashSet<String> {
+    let mut args = vec![
+        "tree",
```
Jakob-Naucke, line 75: I know that the test I introduced also just uses tree, but I think `cargo metadata` (like you use below) should be more stable

tests/no_disallowed_crypto.rs
```diff
@@ -0,0 +1,178 @@
+// SPDX-FileCopyrightText: Chirag Rao <crao@redhat.com>
+//
+// SPDX-License-Identifier: MIT
+
+//! Fedora Crypto Policies (<https://gitlab.com/redhat-crypto/fedora-crypto-policies>)
+//! only curates policy for an allowlist of backends (OpenSSL, GnuTLS, NSS, libkrb5, ...). Any crate that reimplements its own crypto/TLS bypasses the system-wide policy and should be blocked.
+
+//! ////////////// Test Summary /////////////////////
+//! Uses `cargo tree -e normal` to find every crate that actually ships in the built binaries (excluding dev-/build-only dependencies, e.g. codegen tools or build-time fetchers, which never run in production), and `cargo metadata` to look up each one's crates.io categories/keywords. Each crypto crate available in Cargo ecosystem is blocked by default, unless allowlisted. This test can run air-gapped as all information about the Cargo crates is available in the metadata and the .lock files.
```
Jakob-Naucke: nit: please wrap these comments

.github/workflows/lint.yml
```diff
@@ -57,3 +57,5 @@ jobs:
         run: cargo test --test equal_conditions
       - name: "Ensure rustls isn't used"
         run: "! { cargo tree -e normal | grep rustls | grep -v rustls-pki-types; }"
+      - name: "Ensure no disallowed crypto crates are shipped"
```
Jakob-Naucke: Can drop the test above, right?

Jakob-Naucke (commented):
Indeed, we should get oci-client to use jsonwebtoken-openssl when native-tls is enabled, but it's a bit of a separate issue.

tests/no_disallowed_crypto.rs
SpaceFace02: Yes, we would need both in this case, `cargo tree` to get workspace members, and `cargo metadata` to get information about the packages in each workspace.

Jakob-Naucke: > Yes, we would need both in this case, `cargo tree` to get workspace members, and `cargo metadata` to get information about the packages in each workspace.

Ah yes, I missed that `cargo metadata` can't filter out dev dependencies (which we want) as per https://github.com/rust-lang/cargo/issues/14794. According to my browser history, I checked the same thing when I introduced the simpler test :see_no_evil:

Jakob-Naucke (approved):

PR #310, opened 2026-07-10
Integration tests: check reboot with boot ID, test Ingress resolvability, wait for image committed

Jakob-Naucke:
- Waiting for SSH unavailability is unreliable when the reboot is fast. Instead, check for SSH availability and a changed boot ID.
- openshift: Test resolvability of ingress and retry up to 3 times. Also bump timeout for creation.
- Wait for ApprovedImage to be Committed to catch early if it wasn't committed, which would lead to Trustee
failures later

First two commits assisted by AI.

## Summary by Sourcery

Improve integration test reliability around VM reboot behavior, OpenShift ingress exposure, and ApprovedImage readiness.

New Features:
- Add DNS resolvability checks with retries when exposing OpenShift services via LoadBalancer.
- Add a boot ID retrieval helper for VM backends to verify reboots in tests.
- Add a test wait step for the ApprovedImage resource to reach the Committed condition before proceeding.

Enhancements:
- Increase OpenShift ingress creation timeout and introduce multiple exposure attempts with fallback to ClusterIP.
- Replace SSH unavailability-based reboot detection with boot ID change verification in VM reboot-related tests.
- Rename the trustee ConfigMap reference in trusted execution cluster tests to use the trustee-rv-data name.

Tests:
- Tighten integration tests to depend on ApprovedImage being committed and on verified VM reboot via boot ID rather than transient SSH failures.

### DIFF
--- a/test_utils/src/lib.rs
+++ b/test_utils/src/lib.rs
@@ -10,6 +10,7 @@ use k8s_openapi::api::core::v1::{
     ConfigMap, LoadBalancerStatus, Namespace, Secret, Service, ServicePort, ServiceSpec,
     ServiceStatus,
 };
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::Condition;
 use kube::api::{DeleteParams, ObjectMeta, Patch};
 use kube::runtime::wait::await_condition;
 use kube::{Api, Client};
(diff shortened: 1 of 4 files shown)


alicefr (approved):

alicefr:
Looks good and good idea about the boot id!

PR #314, opened 2026-07-16
dependabot: Fix blocking k8s 1.36

Jakob-Naucke:
- Block version we don't want to upgrade to, not the one we're already on or Dependabot ignores it as seen on #311
- Changed the comment: Blocker is no longer RHEL, but still OpenShift CI

## Summary by Sourcery

Build:
- Update Dependabot ignore rules to block k8s.io and sigs.k8s.io packages starting from newer major/minor versions aligned with current Kubernetes 1.36 concerns.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,15 +27,15 @@ updates:
         patterns:
           - "*"
     ignore:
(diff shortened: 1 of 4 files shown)


Jakob-Naucke:
drive-by-adding a small docs fix

alicefr (approved):

Issue #281, opened 2026-06-16
Use dependabot for Rust toolchain updates


Jakob-Naucke:
Dependabot can [update](https://github.blog/changelog/2025-08-19-dependabot-now-supports-rust-toolchain-updates/) the Rust toolchain, but we have the additional requirement that we only move as far as UBI. Investigate if this is also possible.

PR #317, opened 2026-07-23
Release tooling to automate releases and GHA pipeline to build operat…

### LINKED ISSUE #261: Create GHA for releasing the operator images and manifest
Automate the images and manifests tagging,publication when the release is cut

### LINKED ISSUE #286: Fix the versioning of the component
Currently, the version of the component is set to [`0.2.0`](https://github.com/trusted-execution-clusters/operator/blob/main/operator/src/main.rs#L38) which is wrong since we have already the release 0.2.1. We should set this as default equal to latest and add an option to the command line to pass the version of the component or to overwrite it at build time

SpaceFace02:
This PR aims to improve the release workflow of the operator, by introducing helper scripts, GHA pipeline to build and push operator images and other release improvements.

This PR also bumps up Operator version to `0.2.2`, which is our first release publicly available on Github.

This includes commits that:
1. Track operator-sdk version in a seperate go.mod file nested within `tools/`, making it easier for dependabot to create update PRs.
2. All release tooling required to bump versions across all files and GHA workflow to build and push images to Quay.
3. One commit bumping all versions to 0.2.2, which serves as first release being available on Github.

## Summary by Sourcery

Automate the operator release process by introducing version-aware defaults, a prepare-release script, and a GitHub Actions workflow that builds and publishes images and OLM bundles from tagged releases.

New Features:
- Add a GitHub Actions release workflow that builds and pushes operator and component images plus the OLM bundle on semantic version tags.
- Introduce a prepare-release Makefile target and script to consistently bump crate, CSV, README, and lockfile versions for a new release.

Enhancements:
- Make operator image and component versions configurable via environment, defaulting to the trustee version and the Cargo package version.
- Adjust bundle generation to derive the OLM bundle version from the image tag by stripping a leading 'v', and clarify this behavior in the README.
- Update CSV and related image references to the v0.2.1 release of the operator and its managed components.

CI:
- Add a release GitHub Actions workflow that runs on semantic version tags, installs required tooling, logs into Quay, and runs the release build and push pipeline.

Documentation:
- Refresh README release instructions and examples to use v-prefixed image tags and explain OLM semver requirements.


SpaceFace02:
Aims to fix #286 and even #261, and other enhancements.

.github/workflows/release.yml
```diff
@@ -0,0 +1,65 @@
+# SPDX-FileCopyrightText: Chirag Rao <crao@redhat.com>
+#
+# SPDX-License-Identifier: CC0-1.0
+
+name: "Release"
+on:
+  push:
+    tags:
```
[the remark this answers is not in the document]
SpaceFace02: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet

Github does allow regex style matching, its a false positive I believe.

SpaceFace02:
Q: Do we have quay secrets configured for GHA?

.github/workflows/release.yml
```diff
@@ -0,0 +1,65 @@
...
+        with:
+          go-version-file: go.mod
+
+      # generate-bundle-prod.sh invokes `yq` directly on PATH.
+      - name: "Install yq"
+        run: go install github.com/mikefarah/yq/v4@$(go list -m -f '{{.Version}}' github.com/mikefarah/yq/v4)
+
+      - name: "Install operator-sdk"
+        run: |
+          OPERATOR_SDK_VERSION=v1.39.1
```
Jakob-Naucke: Can you track this tool in `tools/` and add that to `.github/dependabot.yml` like with virtctl?

bundle/static/manifests/trusted-cluster-operator.clusterserviceversion.yaml
```diff
@@ -39,9 +39,9 @@ metadata:
           },
           "spec": {
             "trusteeImage": "quay.io/trusted-execution-clusters/key-broker-service:v0.17.0",
-            "pcrsComputeImage": "quay.io/trusted-execution-clusters/compute-pcrs:0.2.0",
-            "registerServerImage": "quay.io/trusted-execution-clusters/registration-server:0.2.0",
-            "attestationKeyRegisterImage": "quay.io/trusted-execution-clusters/attestation-key-register:0.2.0",
+            "pcrsComputeImage": "quay.io/trusted-execution-clusters/compute-pcrs:v0.2.1",
```
Jakob-Naucke: good catch, can this be a separate commit?

operator/src/main.rs
```diff
@@ -32,10 +32,17 @@ mod trustee;
 use crate::conditions::*;
 use operator::*;
 
-/// Default tag for Trustee image
-const TRUSTEE_VERSION: &str = "v0.17.0";
```
Jakob-Naucke, line 36: The single source of truth of which version to use should be `RELATED_IMAGE_*`

scripts/prepare-release.sh
```diff
@@ -0,0 +1,138 @@
...
+# Steps followed:
+# 1. Validate passed params.
+# 2. Update [package] version in all related Cargo.toml files in our owned workspaces.
+# 3. Update CSV template with new version.
+# 4. Update README with new version.
+# 5. Refresh Cargo.lock for workspace package versions.
+
+set -euo pipefail
+
+DRY_RUN=false
```
Jakob-Naucke, line 19: I don't really get what the dry run is for -- if it's enabled, you get a bunch of `echo`s but I can't tell how it would tell its user what they set up wrong (and exits on an incorrectly formatted version or `cargo metadata` bailing out you also get without the dry run)

SpaceFace02: Yeah, in its current state I admit it might not be very useful as is, as I didn't spend much time on the dry-run part of the script, as its just a few `echos` to get the overall picture of what the script is doing. I could either remove it, or try to add version resolution and more information into the `echos`. (I'll lean into the latter).

(Btw the dry-mode inspiration came from how the trustee operator on how they are doing releases, ([here](https://github.com/confidential-containers/trustee-operator/blob/main/hack/release/do-release.sh)). so decided to add it in.

operator/src/main.rs
SpaceFace02: Just to be clear, this is the fallback chain we are following right?

1. CR spec (if user provides an override)
2. `RELATED_IMAGE_*` environment variable
3. Hardcoded default (for standalone/dev deployments)

.github/workflows/release.yml
SpaceFace02: Hmm, yeah that makes it easier for dependabot, but might increase go.mod size and bloat it up with its extra transitive dependencies.

I don't think it might be needed becauses releases are much less frequent than integration test runs (which warrants tracking virtctl).

WDYT?

bundle/static/manifests/trusted-cluster-operator.clusterserviceversion.yaml
SpaceFace02: You mean a (seperate commit for bundle bumps), or  a (seperate commit for release tooling and another commit for the actual bumps across all files)?

.github/workflows/release.yml
Jakob-Naucke: While it is a lot of LOC, it's not in the project root's `/go.mod` but nicely tucked away in `/tools/operator-sdk/go.mod`. I think we want to auto update whenever we can, and dependabot lets us.

operator/src/main.rs
Jakob-Naucke: 1 is no more, 2 & 3 are correct.

bundle/static/manifests/trusted-cluster-operator.clusterserviceversion.yaml
Jakob-Naucke: I think the former is enough.

scripts/prepare-release.sh
Jakob-Naucke: Let's do the latter, and at least also print e.g. the `$VERSION`

SpaceFace02:
Whoops, looks like adding operator-sdk in tools.go.mod pulled in all these known CVE's of these transitive dependencies, I'm reconsidering just using a pre-built binary and manually bumping the version.

WDYT? False alarm because go.mod is seperate, and tucked inside tools, or something that can increase attack surface?

bundle/static/manifests/trusted-cluster-operator.clusterserviceversion.yaml
SpaceFace02: I split it across 3 commits to make it easier for bisects and reviews:
1. Changes to `/tools` to track operator-sdk.
2. Release tooling changes.
3. Actual version bumps across all files. 

Bundle version bumps going in a seperate commit may break bisects, furthermore we will have 2 commits that are very tightly coupled to each other. Hence I followed the above strategy.

Jakob-Naucke:
> WDYT? False alarm because go.mod is seperate, and tucked inside tools, or something that can increase attack surface?

can you bump from 1.42.0 to 1.42.3? it seems they already picked up many of the mitigations to the CVEs listed here

SpaceFace02:
> can you bump from 1.42.0 to 1.42.3? it seems they already picked up many of the mitigations to the CVEs listed here

1.42.3 requires go version 1.26.3, but we are currently using 1.25.0 for our operator and related go.mod files.

https://github.com/trusted-execution-clusters/operator/blob/main/.github/dependabot.yml#L29

https://github.com/operator-framework/operator-sdk/compare/v1.42.1...v1.42.3

1.42.1 is the highest we can go if we are constrained to be under go version 1.26.0

SpaceFace02:
> 1.42.3 requires go version 1.26.3, but we are currently using 1.25.0 for our operator and related go.mod files.

Although imo we can go ahead with version 1.26.3 for the `/tools/operator-sdk/go.mod` file as operator-sdk go.mod is in no way related to actual operator functionality, and is used only when we do releases.

Jakob-Naucke:
> > 1.42.3 requires go version 1.26.3, but we are currently using 1.25.0 for our operator and related go.mod files.
> 
> Although imo we can go ahead with version 1.26.3 for the `/tools/operator-sdk/go.mod` file as operator-sdk go.mod is in no way related to actual operator functionality, and is used only when we do releases.

Yeah I agree, the restriction is really just for the OpenShift CI but it won't touch that

Jakob-Naucke (commented):
nit: wouldn't put the date in the release commit

operator/src/main.rs
```diff
@@ -32,10 +32,15 @@ mod trustee;
 use crate::conditions::*;
 use operator::*;
 
-/// Default tag for Trustee image
+// Default fallback version tag for Trustee image if RELATED_IMAGE_TRUSTEE is not set.
```
Jakob-Naucke: nit: let's keep this a doc comment (`///`)

Jakob-Naucke (approved):
one nit but LGTM otherwise. would like @alicefr's or @uril's review though because I haven't performed a release.

Jakob-Naucke:
Didn't notice the failing builds. IMO we could bump buildroot to Fedora 44.

SpaceFace02:
> Didn't notice the failing builds. IMO we could bump buildroot to Fedora 44.

Idk bumping the buildroot will help because the error isn't inside the buildroot container, I think its in the GHA host running `make crds-rs` [here](https://github.com/trusted-execution-clusters/operator/blob/main/Makefile#L57), as its scanning all directories.

The go version mismatch is causing the build to fail, because the CI expects version 1.25.0 everywhere, but `tools/operator-sdk` is at 1.26.3 due to our constraints.

I can skip running generate on tools directory for now

SpaceFace02:
Unrelated to this PR, but this issue (#[281](https://github.com/trusted-execution-clusters/operator/issues/281)), we can only go as far as UBI, which effectively means as far as RHEL (v10?). These constraints are mainly for Openshift CI I believe. Are we running Openshift CI air-gapped? So we can't install or update toolchains when needed? Are we using package managers to install rust/go, or are we using toolchain managers like rustup?

.github/workflows/release.yml
```diff
@@ -0,0 +1,65 @@
...
+        with:
+          go-version-file: go.mod
+
+      # generate-bundle-prod.sh invokes `yq` directly on PATH.
+      - name: "Install yq"
+        run: go install github.com/mikefarah/yq/v4@$(go list -m -f '{{.Version}}' github.com/mikefarah/yq/v4)
+
+      - name: "Install operator-sdk"
+        run: |
+          version=$(awk '$1 == "require" && /operator-framework\/operator-sdk/ {print $3}' tools/operator-sdk/go.mod)
```
Jakob-Naucke, line 51: Seems legit, maybe add a comment on why reading the version with `go` is unadvisable before someone tries to "fix" it

Jakob-Naucke (commented):
> Unrelated to this PR, but this issue (#https://github.com/trusted-execution-clusters/operator/issues/281), we can only go as far as UBI, which effectively means as far as RHEL (v10?). These constraints are mainly for Openshift CI I believe. Are we running Openshift CI air-gapped? So we can't install or update toolchains when needed? Are we using package managers to install rust/go, or are we using toolchain managers like rustup?

UBI has 1.26, including on 9. However, the base images on OpenShift CI only go up to 1.25 as of today in accordance with what's required for the respective Kubernetes versions. They are not airgapped, but I'd rather run what they provide unless we have a very good reason not to. For Rust, we indeed use Rustup there.

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+        uses: actions/checkout@v7
+
+      - name: "Install Go toolchain"
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
+      # generate-bundle-prod.sh invokes `yq` directly on PATH.
+      - name: "Install yq"
+        run: go install github.com/mikefarah/yq/v4@$(go list -m -f '{{.Version}}' github.com/mikefarah/yq/v4)
```
uril: This line already exists in the Makefile.
I'm not sure if it's better here to 'make yq' or keep this simple line for installation.

tools/operator-sdk/go.mod
```diff
@@ -0,0 +1,247 @@
...
+//
+// SPDX-License-Identifier: CC0-1.0
+
+module io/trustedexecutioncluster/tools/operator-sdk
+
+go 1.26.3
+
+require github.com/operator-framework/operator-sdk v1.42.3
+
+require (
```
uril, line 11: I think this require can be removed as we only care about operator-sdk itself

uril:
Indirectly related: can we do the same for yq, kopium and controller-gen - is there a need to build them or can their pre-built binaries be downloaded?

uril:
As part of this PR you actually create a 0.2.2 release.
Please either do it in a separate PR such that it's clear or mention it in the PR description

It would be nice (but not currently necessary) to summarize what's changed since the previous release.
Update: As part of the 'release' github action, release-notes are being generated.

scripts/prepare-release.sh
```diff
@@ -0,0 +1,190 @@
...
+done
+echo "  - $CSV"
+echo "  - $README"
+echo "  - Cargo.lock"
+echo ""
+echo "Next steps:"
+echo "  1. Review changes: git diff"
+echo "  2. Commit: git commit -am 'Release $VERSION'"
+echo "  3. Tag: git tag $VERSION"
+echo "  4. Push: git push origin main $VERSION"
```
uril: After steps 1,2 - create a PR
After the PR is accepted:
- 3 should be done (can be automatically via the 'release' github action)
- 4 is not needed as it is automatically done by the PR being accepted.

scripts/prepare-release.sh
```diff
@@ -0,0 +1,190 @@
...
+    echo "ERROR: VERSION must not start with 'v'. Got: $VERSION"
+    exit 1
+fi
+
+# Require MAJOR.MINOR.PATCH (semver).
+if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+    echo "ERROR: VERSION must match MAJOR.MINOR.PATCH. Got: $VERSION"
+    exit 1
+fi
+
```
uril, line 53: Do we want to also check the new release is an increment of the current one (we don't have to) ?
If MAJOR or MINOR are incremented, then MINOR and/or PATCH reset back to 0.

uril:
This PR will help us with cutting a release. Thanks @SpaceFace02

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+  cancel-in-progress: false
+
+env:
+  CONTAINER_CLI: docker
+  BUILD_TYPE: release
+  REGISTRY: quay.io/trusted-execution-clusters
+
+jobs:
+  release:
+    name: "Build and push images"
```
uril, line 25: This can later be moved to a build_and_push_images github action (and called here) such that occasionally (nightly/weekly) 'latest' images can be built too.

SpaceFace02: Yes, that is a good idea. We would also need to decide the frequency of releases before we go GA and have a bigger community.

.github/workflows/release.yml
SpaceFace02: The one in the Makefile has renamed the yq binary with a version suffix, like `yq-4.4.4`, and accesses it like `$(YQ)`. Furthermore, the generate-bundle-prod.sh file uses yq as is, without the versioning suffix. I think we can keep this simple one line for now.

scripts/prepare-release.sh
SpaceFace02: Yes, that's a good point, I have addressed it in the latest commit. I have added a check which only allows specific versions to be the next version, according to the current version to avoid human error while tagging.

scripts/prepare-release.sh
SpaceFace02: Thanks for pointing this out, I have clarified the workflow a bit better.

tools/operator-sdk/go.mod
SpaceFace02: I believe these are the indirect dependencies of operator-sdk, which are pulled in by go, when running `go mod tidy`. I have just added the one on line 9.

SpaceFace02 (commented):
Thanks for the review @uril! I have addressed the comments

SpaceFace02:
> Indirectly related: can we do the same for yq, kopium and controller-gen - is there a need to build them or can their pre-built binaries be downloaded?

Yes, all 3 (yq, kopium and controller-gen) have release cycles and pre built binaries. Currently the integration tests in the CI are using `make crds-rs` to generate the crds, using these tools which we build ourselves. We could modify the make file to download the appropriate binaries in the future.

> As part of this PR you actually create a 0.2.2 release. Please either do it in a separate PR such that it's clear or mention it in the PR description
> 
> It would be nice (but not currently necessary) to summarize what's changed since the previous release. Update: As part of the 'release' github action, release-notes are being generated.

Since this is our first release, [this line](https://github.com/trusted-execution-clusters/operator/pull/317/changes#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R66): makes sure there are release notes, which include all commits from the start. We would need to cut the description and trim down the release notes manually, using AI to summarize better.

I will update the PR description, I have a seperate commit for the version bump and release.

SpaceFace02:
I believe we can merge this then? Is there anything that needs to be done?

SpaceFace02:
> Indirectly related: can we do the same for yq, kopium and controller-gen - is there a need to build them or can their pre-built binaries be downloaded?

@uril I have merged [this PR](https://github.com/trusted-execution-clusters/operator/pull/326), which addresses this issue and uses pre built components whenever it can, and falls back to `go install` if the URLS change or there is any transient error.

PR #318, opened 2026-07-23
Unified Container Build with Cache Mounts

yairpod:
Until now our builds have compiled the same files 4 times, As the builds were done in separate containers and had no access to the compilation objects from the other containers or previous builds. By Unifying the Build stage into one Container and interceding build caches we will prevent unnecessary work and achieve faster builds.

To check the effectiveness of these changes I used Opus 4.6 to write a build time benchmark script (see attached) and run a short test (also see attached).
The Highlights are these:
Without this PR the Operator container build was up to 30% faster the second build to the first build (~220 second time ~310 first time) , with the smaller gains for the other containers.
With the PR the difference is 70% (~110 second time ~305 first time) , but the other containers show a much larger improvement, as the speed up is about 90% (~30 seconds first build to ~3 in second build), also the build time of the first build of the other containers is reduced from the original code build time of ~220 seconds to ~30 in united build, is is due to the containers being build togather.

[benchmark-builds.sh](https://github.com/user-attachments/files/30302940/benchmark-builds.sh)
[results.csv](https://github.com/user-attachments/files/30302945/results.csv)
[summary.txt](https://github.com/user-attachments/files/30302946/summary.txt)

## Summary by Sourcery

Unify Rust binary container builds into a single cached builder stage and derive multiple runtime images from it to reduce duplicate compilation and improve build performance.

New Features:
- Introduce multi-stage Containerfile targets for operator, compute-pcrs, register-server, and attestation-key-register binaries.
- Add integration of the reference-values Git repository into the build to bundle required reference data into compute-pcrs images.

Enhancements:
- Enable shared cargo build caches and a unified cargo build invocation to accelerate debug and release builds across all binaries.
- Refine Cargo workspace membership and dev-dependency handling to support unified builds and dependency pre-compilation.
- Update Makefile targets to build all images from the unified Containerfile, and add container builder cache pruning to the clean target.

Build:
- Set a default build_type argument and introduce multi-stage, target-based container builds driven from a single Containerfile while removing per-binary Containerfiles.

Tests:
- Wire the reference-values crate into the tests workspace for consistent usage in test binaries.


Makefile
```diff
@@ -101,13 +101,13 @@ CONTAINER_CLI ?= podman
...
-	$(CONTAINER_CLI) build $(IMAGE_BUILD_OPTIONS) -t $(COMPUTE_PCRS_IMAGE) -f compute-pcrs/Containerfile .
+	$(CONTAINER_CLI) build $(IMAGE_BUILD_OPTIONS) --target compute-pcrs -t $(COMPUTE_PCRS_IMAGE) -f Containerfile .
 reg-server-image:
-	$(CONTAINER_CLI) build $(IMAGE_BUILD_OPTIONS) -t $(REG_SERVER_IMAGE) -f register-server/Containerfile .
+	$(CONTAINER_CLI) build $(IMAGE_BUILD_OPTIONS) --target register-server -t $(REG_SERVER_IMAGE) -f Containerfile .
 attestation-key-register-image:
-	$(CONTAINER_CLI) build $(IMAGE_BUILD_OPTIONS) -t $(ATTESTATION_KEY_REGISTER_IMAGE) -f attestation-key-register/Containerfile .
+	$(CONTAINER_CLI) build $(IMAGE_BUILD_OPTIONS) --target attestation-key-register -t $(ATTESTATION_KEY_REGISTER_IMAGE) -f Containerfile .
 
 image: operator-image compute-pcrs-image reg-server-image attestation-key-register-image
```
Jakob-Naucke: These were written with parallelization in mind, but building like this makes more sense in sequence. Honestly I'd maybe just go back to one target & sequence

Containerfile
```diff
@@ -1,34 +1,79 @@
...
+COPY attestation-key-register/Cargo.toml attestation-key-register/
+COPY attestation-key-register/src/lib.rs attestation-key-register/src/
 
-# Set only required crates as members to minimize rebuilds upon changes.
-RUN sed -i 's/members = .*/members = ["lib", "operator"]/' Cargo.toml && \
+RUN sed -i 's/members = .*/members = ["lib", "operator", "compute-pcrs", "register-server", "attestation-key-register"]/' Cargo.toml && \
     sed -i '/\[dev-dependencies\]/,$d' operator/Cargo.toml && \
+    sed -i '/\[dev-dependencies\]/,$d' register-server/Cargo.toml && \
     sed -i '/trusted-cluster-operator-test-utils/d' lib/Cargo.toml && \
+    git clone --depth 1 https://github.com/trusted-execution-clusters/reference-values && \
```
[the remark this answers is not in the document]
Jakob-Naucke: Yes, but I think it should be tracked as a Cargo dependency à la

```toml
[dev-dependencies]
reference-values = { git = "https://github.com/trusted-execution-clusters/reference-values" }
```

and then we get updates via Dependabot, and then here you'd do

```suggestion
    rv_source=$(cargo metadata --format-version 1 | \
        jq -r '.packages[] | select(.name == "reference-values") | .source') && \
    rv_repo=$(echo "$rv_source" | sed 's/^git+//;s/[?#].*//') && \
    rv_commit=$(echo "$rv_source" | sed 's/.*#//') && \
    git clone --depth 1 --revision "$rv_commit" "$rv_repo" reference-values && \
```

This depends on https://github.com/trusted-execution-clusters/reference-values/pull/9.

Jakob-Naucke (commented):
very neat with the `.containerignore` also

uril:
1. I assume that having 4 separate Containerfiles, but with cache would result in a similar improvement (but needs to be benchmarked).
2. Please make sure that make clean also cleans the cache.

Containerfile
```diff
@@ -1,34 +1,83 @@
...
+COPY compute-pcrs/src compute-pcrs/src
+COPY register-server/src register-server/src
+COPY attestation-key-register/src attestation-key-register/src
 
-# Distribution stage
-FROM quay.io/fedora/fedora:43
-ARG build_type
-COPY --from=builder "/build/target/$build_type/operator" /usr/bin
+RUN --mount=type=cache,target=/build/target \
+    --mount=type=cache,target=/usr/local/cargo/registry \
```
uril, line 60: I am not sure  /usr/local/cargo/registry is the right directory -- shouldn't it be /root/.cargo/registry ?

Containerfile
```diff
@@ -1,34 +1,83 @@
...
+    mkdir -p /output && \
+    cp /build/target/${profile_dir}/operator /output/ && \
+    cp /build/target/${profile_dir}/compute-pcrs /output/ && \
+    cp /build/target/${profile_dir}/register-server /output/ && \
+    cp /build/target/${profile_dir}/attestation-key-register /output/ && \
+    mkdir -p /output/reference-values && \
+    cp -r /build/reference-values/efivars /output/reference-values/ && \
+    cp -r /build/reference-values/mok-variables /output/reference-values/
+
+# Distribution stages
```
SpaceFace02, line 83: Doesn't need to go in this PR, but we could investigate using a lighter distribution image for our final build stage.

Raised #321 for the same.

Containerfile
yairpod: Interesting, seems the path is configurable, I tried adding the path you suggested but the benchmark remained unchanged.
[bench] [1/6] new | debug | rep 1 (cold)
[bench]   -> 378.8s (operator=279.3s compute-pcrs=31.4s register-server=34.7s attestation-key-register=33.4s)
[bench] [2/6] new | debug | rep 2 (warm)
[bench]   -> 123.1s (operator=111.0s compute-pcrs=4.8s register-server=3.5s attestation-key-register=3.8s)
[bench] [3/6] new | debug | rep 3 (warm)    
[bench]   -> 119.4s (operator=107.7s compute-pcrs=4.5s register-server=3.6s attestation-key-register=3.6s)

uril: It may be unused - can you please try without this line ?

yairpod: After some testing and comparison i found that without at least one of these paths (/usr/local/cargo/registry or/root/.cargo/registry), the caching is not working as well (second build takes ~220 seconds instead of ~120).
From what I understand the caching directory is configurable, and can be either one of these, but we do need at least one.
So I am would suggest we keep the original.

Containerfile
yairpod: Good Idea, we should look into it, but as you wrote, it's a separate issue and PR

yairpod:
> 1. I assume that having 4 separate Containerfiles, but with cache would result in a similar improvement (but needs to be benchmarked).
I made the variation with 4 separate Containerfiles that have build caches, and benchmarked it (see results attached), as you can see, the build time for the first container is not significantly impacted but the next 3 lose much of the improvement.
[split_summary.txt](https://github.com/user-attachments/files/30652058/split_summary.txt)


>     2. Please make sure that make clean also cleans the cache.
Added cache cleaning to "make clean" command.

Makefile
yairpod: We can do that, but would that impact CI and testing infrastructure?

Jakob-Naucke:
@yairpod is this PR closed intentionally?

yairpod:
> @yairpod is this PR closed intentionally?

No, thanks for pointing this out

Makefile
Jakob-Naucke: no (and IMO the CI should follow our needs and not vice versa whenever possible)

yairpod: Done

Jakob-Naucke (approved):

Jakob-Naucke:
ah yes, the verified signatures

Makefile
```diff
@@ -185,6 +180,7 @@ clean:
 	cargo clean
 	rm -rf bin manifests $(CRD_YAML_PATH) $(CRD_RS_PATH)
 	rm -f trusted-cluster-gen config/rbac/role.yaml .crates.toml .crates2.json
+	$(CONTAINER_CLI) builder prune --all --force --filter label=project=trusted-cluster-operator
```
uril, line 199: should "builder" be replaced with -- image , system, volume, other ?

Makefile
```diff
@@ -100,16 +100,11 @@ cluster-down:
...
-compute-pcrs-image:
-	$(CONTAINER_CLI) build $(IMAGE_BUILD_OPTIONS) -t $(COMPUTE_PCRS_IMAGE) -f compute-pcrs/Containerfile .
-reg-server-image:
-	$(CONTAINER_CLI) build $(IMAGE_BUILD_OPTIONS) -t $(REG_SERVER_IMAGE) -f register-server/Containerfile .
-attestation-key-register-image:
-	$(CONTAINER_CLI) build $(IMAGE_BUILD_OPTIONS) -t $(ATTESTATION_KEY_REGISTER_IMAGE) -f attestation-key-register/Containerfile .
-
-image: operator-image compute-pcrs-image reg-server-image attestation-key-register-image
+image:
+	$(CONTAINER_CLI) build $(IMAGE_BUILD_OPTIONS) --target operator -t $(OPERATOR_IMAGE) -f Containerfile .
```
uril, line 120: What's the purpose of this change -- to always build all 4 containers ?

Jakob-Naucke: To avoid performing identical steps in parallel with multithreaded make. If we want to keep the targets, one could also do
```suggestion
	$(MAKE) operator-image
	$(MAKE) compute-pcrs-image
	$(MAKE) reg-server-image
	$(MAKE) attestation-key-register-image
```

yairpod: Also, the caches are behind a lock, so multithreading is sequential anyway.

Jakob-Naucke: oh, but when on your previous revision 108c2b0b, `make -j4 image` yields something like
```
[1/5] STEP 17/23: RUN make crds-rs
--> 1a27730f3d14
[1/3] STEP 16/23: RUN sed -i 's/members = .*/members = ["lib", "operator", "compute-pcrs", "register-server", "attestation-key-register"]/' Cargo.toml &&     sed -i '/\[dev-dependencies\]/,$d' operator/Cargo.toml &&     sed -i '/\[dev-dependencies\]/,$d' register-server/Cargo.toml &&     sed -i '/trusted-cluster-operator-test-utils/d' lib/Cargo.toml
--> 376893283ee7
[1/4] STEP 17/23: RUN make crds-rs
--> 33d731781ac3
[1/2] STEP 17/23: RUN make crds-rs
    Updating crates.io index
    Updating git repository `https://github.com/trusted-execution-clusters/compute-pcrs`
--> 4d4aca09ee8d
[1/3] STEP 17/23: RUN make crds-rs
    Updating crates.io index
    Updating git repository `https://github.com/trusted-execution-clusters/compute-pcrs`
    Updating crates.io index
    Updating git repository `https://github.com/trusted-execution-clusters/compute-pcrs`
    Updating git repository `https://github.com/latchset/clevis-pin-trustee`
    Updating git repository `https://github.com/latchset/clevis-pin-trustee`
    Updating crates.io index
    Updating git repository `https://github.com/trusted-execution-clusters/compute-pcrs`
    Updating git repository `https://github.com/latchset/clevis-pin-trustee`
    Updating git repository `https://github.com/latchset/clevis-pin-trustee`
 Downloading crates ...
 Downloading crates ...
```

which doesn't look so sequential to me?

yairpod: I'm not sure, I read about the Lock and did not experiment with it.
But that seems to be parallelization of line 40 "RUN make crds-rs" and the first use of --mount=type=cache (where the lock should be) is line 43 (in revision [108c2b0](https://github.com/trusted-execution-clusters/operator/commit/108c2b0b4da86f8edf451db7ec0ac0ee357a8de9)) .
Also that is a huge waste of build effort as a sequential build would not run these stages 4 times, the layers would be pulled from the intermediate layer cache (not to be confused with the mount cache).

Makefile
yairpod: builder prune is different meaning the the others:
builder prune - Removes BuildKit build cache (intermediate layers) 
image prune - Removes Dangling/unused images
system prune - Removes Everything (containers, images, networks, cache)
volume prune - Removes Unused volumes

So we need to clean the BuildKit build cache, but I just found that it does not clean the mount caches.
I will add the command "$(CONTAINER_CLI) builder prune --force --filter type=exec.cachemount" for that, note that filter and labels are not supported here! so we will have to clean all mount caches in the system, or none. I think that in this case cleaning all of them should be acceptable.

Makefile
yairpod: I added mount caches to the make crds-rs stage, it shows improvement in build times when the libs are modified.


Benchmarks:
_New is with make crds-rs caches, old is the PR without them [60dff635](https://github.com/trusted-execution-clusters/operator/commit/60dff635cc763e7df989fb3a42cc37a1d21715a2)_

approach  build_type  repetition  cache_state  edit_target  total_time_s
new       debug       1           cold         none         402.5
new       debug       2           warm         lib          92.5
new       debug       3           warm         lib          93.1
new       debug       2           warm         operator     22.5
new       debug       3           warm         operator     23.3
old       debug       1           cold         none         404.0
old       debug       2           warm         lib          114.0
old       debug       3           warm         lib          119.4
old       debug       2           warm         operator     22.5
old       debug       3           warm         operator     23.3

Containerfile
```diff
@@ -1,34 +1,99 @@
...
 COPY operator/Cargo.toml operator/
 COPY operator/src/lib.rs operator/src/
+COPY compute-pcrs/Cargo.toml compute-pcrs/
+COPY compute-pcrs/src/lib.rs compute-pcrs/src/
+COPY register-server/Cargo.toml register-server/
+COPY register-server/src/lib.rs register-server/src/
+COPY attestation-key-register/Cargo.toml attestation-key-register/
+COPY attestation-key-register/src/lib.rs attestation-key-register/src/
 
-# Set only required crates as members to minimize rebuilds upon changes.
```
alicefr, line 18: nit: is the comment not true any longer?

Containerfile
```diff
@@ -1,34 +1,99 @@
...
+
+# Unified builder stage — compiles all binaries in a single cargo invocation.
 FROM ghcr.io/trusted-execution-clusters/buildroot:fedora AS builder
+LABEL project=trusted-cluster-operator
 ARG build_type
 WORKDIR /build
 
 COPY Makefile Cargo.toml Cargo.lock go.mod go.sum .
+
+RUN rv_line=$(grep -A3 'name = "reference-values"' Cargo.lock | grep 'source = ') && \
```
alicefr, line 17: Do we have a more robust way to parse the repo and the commit instead of counting the line after the grep? What if it changes?

tests/Cargo.toml
```diff
@@ -23,6 +23,7 @@ serde_json.workspace = true
 tokio = { workspace = true }
 trusted-cluster-operator-lib = { path = "../lib" }
 trusted-cluster-operator-test-utils = { path = "../test_utils" }
+reference-values.workspace = true
```
alicefr, line 26: Is this used? It wasn't present before

Containerfile
```diff
@@ -1,34 +1,99 @@
...
+    cp /build/target/${profile_dir}/register-server /output/ && \
+    cp /build/target/${profile_dir}/attestation-key-register /output/
+
+RUN mkdir -p /output/reference-values && \
+    cp -r /build/reference-values/efivars /output/reference-values/ && \
+    cp -r /build/reference-values/mok-variables /output/reference-values/
+
+# Distribution stages
+FROM quay.io/fedora/fedora:43 AS operator
+COPY --from=builder /output/operator /usr/bin
```
alicefr, line 85: can you please add the entrypoint for each new image to be consistent

Containerfile
```diff
@@ -1,34 +1,99 @@
...
+
+FROM quay.io/fedora/fedora:43 AS compute-pcrs
+COPY --from=builder /output/compute-pcrs /usr/bin
+COPY --from=builder /output/reference-values /reference-values
```
alicefr, line 89: Same

Makefile
```diff
@@ -201,6 +196,8 @@ clean:
 	cargo clean
 	rm -rf bin manifests $(CRD_YAML_PATH) $(CRD_RS_PATH)
 	rm -f trusted-cluster-gen config/rbac/role.yaml .crates.toml .crates2.json
+	$(CONTAINER_CLI) builder prune --all --force --filter label=project=trusted-cluster-operator
+	$(CONTAINER_CLI) builder prune --force --filter type=exec.cachemount
```
alicefr, line 200: Don't we need a label for the clean up in this case as well?

alicefr:
@yairpod thanks for the work! Overall, it looks good, just minor stuff

Containerfile
uril: cargo metadata --format-version=1  | jq '.packages.[] | select ( .name == "reference-values" ) | .source'

Makefile
uril: It seems (running with --help) "podman builder prune" is the same as "podman image prune".
I suspected "builder" as "man podman" does not show this command and there is no man "podman-builder". 
We may want to limit it this command with '--build-cache' option

Makefile
yairpod: Unfortunately type=exec.cachemount does not support labels, 
BuildKit cache mounts (the --mount=type=cache in RUN instructions) are global, anonymous volumes — they don't carry image labels.

Containerfile
yairpod: It changed quite a bit, before the section was limiting the Cargo.toml members to include only the dependencies needed by this container, now we are building all 4 containers together, and the sed command basically excludes the testing dependencies ("test_utils" "tests").
We can put a new comment, but it should be somewhat different.

Containerfile
yairpod: Using cargo metadata could be a problem, I tries to use the command you provided and it added much work to build the cargo metadata, but did not return any results for "reference-values".
We could use:
rv_line=$(awk '/^\[\[package\]\]/{p=0} /^name = "reference-values"/{p=1} p && /^source = /{print; exit}' Cargo.lock)
It's not as well connected to the cargo as metadata but more robust then sed.

Containerfile
yairpod: Added

Containerfile
yairpod: Added

tests/Cargo.toml
yairpod: This is used to include the reference-values git commit in the Cargo.lock
However it belongs in compute-pcrs/Cargo.toml not tests/Cargo.toml, and has been moved

Containerfile
Jakob-Naucke: @yairpod the command that @uril sent works for me (plus maybe some `sed` to take the hash from the URL), are you sure you were on the right branch?

yairpod: @Jakob-Naucke  Right branch, might be en env or script change issue, please send me the exact Containerfile you used

Jakob-Naucke:
```Dockerfile
RUN rv_line=$(cargo metadata --format-version=1 | jq '.packages.[] | select ( .name == "reference-values" ) | .source') && echo $rv_line
```

yields

```
  Downloaded k8s-openapi v0.28.0
"git+https://github.com/trusted-execution-clusters/reference-values#342fa08c9bac3253cf972e136e24f7c7c41e9d29"
```

this you'd have to do after the `RUN sed … lib/Cargo.toml` block

Containerfile
```diff
@@ -1,34 +1,104 @@
...
 
-# Set only required crates as members to minimize rebuilds upon changes.
-RUN sed -i 's/members = .*/members = ["lib", "operator"]/' Cargo.toml && \
+RUN sed -i 's/members = .*/members = ["lib", "operator", "compute-pcrs", "register-server", "attestation-key-register"]/' Cargo.toml && \
     sed -i '/\[dev-dependencies\]/,$d' operator/Cargo.toml && \
-    sed -i '/trusted-cluster-operator-test-utils/d' lib/Cargo.toml && \
+    sed -i '/\[dev-dependencies\]/,$d' register-server/Cargo.toml && \
+    sed -i '/trusted-cluster-operator-test-utils/d' lib/Cargo.toml
+
+#RUN rv_line=$(cargo metadata --format-version=1 | jq '.packages.[] | select ( .name == "reference-values" ) | .source') && echo $rv_line && exit 1
```
Jakob-Naucke: this comment can probably be dropped

Makefile
yairpod: Interesting, seems to be different between Docker and Podman.
What I wrote is true only for Docker:
https://docs.docker.com/reference/cli/docker/buildx/prune/#filter
And what you found is only true for Podman:
https://docs.podman.io/en/latest/markdown/podman-image-prune.1.html#build-cache

yairpod: Fixed

Makefile
uril: I'm still a bit concerned about not being able to build a single container image, only all of them  together.
How can the benchmark script measure each of them separately, using make "$target"?

yairpod: I suggest we restore the single container image build options in the makefile,
There are some cases were people want to build just one image (usually the operator), and there is some build time difference.
When the build  of an image is called with the --target flag the builder will skip all stages that are not in it's dependencies, I reworked the stages so the reference-values stages needed only by the compute-pcrs image are separate from the general builder, so now that work is only done when he compute-pcrs image is requested.
Now the build times are shorter for the oporator alone then all images together.
[image_targets_reorg_summary.txt](https://github.com/user-attachments/files/30978789/image_targets_reorg_summary.txt)
[image_targets_reorg_results.csv](https://github.com/user-attachments/files/30978790/image_targets_reorg_results.csv)

uril: Thanks

Containerfile
```diff
@@ -1,34 +1,102 @@
...
     make crds-rs
 
-# In debug builds, build dependencies to avoid full rebuild.
-RUN if [ "$build_type" = debug ]; then cargo build -p operator; fi
+# In debug builds, pre-build dependencies to avoid full rebuild on source changes.
+RUN --mount=type=cache,target=/build/target \
+    --mount=type=cache,target=/usr/local/cargo/registry \
+    if [ "$build_type" = debug ]; then \
+      cargo build -p operator -p compute-pcrs -p register-server -p attestation-key-register; \
+    fi
```
uril, line 46: It would be nice (in a different PR) to run the benchmark with/without this RUN and see if it actually makes the build faster

Jakob-Naucke:
@yairpod FYI the PR won't work on Azure CI because it changes the build structure. we'll have to update there after merge sadly

yairpod:
> @yairpod FYI the PR won't work on Azure CI because it changes the build structure. we'll have to update there after merge sadly

Yes, I was just looking at it,
I think the issue is that the Azure CI is bypassing the Make commands and building the images with it's own Docker build command that goes directly to the old containerfiles. and likely also lacking the --target flags that are now necessary.

I understand the code for these tests is not part of this repo?

Jakob-Naucke:
> Yes, I was just looking at it, I think the issue is that the Azure CI is bypassing the Make commands and building the images with it's own Docker build command that goes directly to the old containerfiles. and likely also lacking the --target flags that are now necessary.

Yes. Pushing images for your tests to use with your own logic isn't really part of OpenShift CI's design AFAICT; instead, you define what Containerfiles (and also targets via [build_args](https://docs.ci.openshift.org/architecture/ci-operator/#build-arguments)) make which image, and that's then built before your test script ever enters.

> I understand the code for these tests is not part of this repo?

No, unfortunately the place where we control this isn't this repo. It's [here](https://github.com/openshift/release/blob/1f1510d8c42ab286fc63242748b20ef67da3ce92/ci-operator/config/trusted-execution-clusters/operator/trusted-execution-clusters-operator-main__azure.yaml#L25)

PR #324, opened 2026-07-30
Use a read-timeout client for non-watch operations

Jakob-Naucke:
Default kube-rs client has no read timeout so that watch operations are supported. Some non-watch operations have been seen to hang. Use a timeout of 30 seconds for such operations.

Introduce a structure to pass both clients where needed.

This yielded all successful runs for many hours on GHA CI and 4 passes on the notoriously flaky OpenShift CI.

Also:
- pass error if Machine cleanup fails
- reconcile ApprovedImages on job changes too for status updates
- use must-gather in GHA
- fix a log message and uses of `warn!` in ak-reg
- add timestamps to logging, which

Fixes: #227

@iroykaufman I also tested this with #248, here's the [rerere cache](https://git-scm.com/book/en/v2/Git-Tools-Rerere) already in case we merge this first: [rerere-timed-client.tar.gz](https://github.com/user-attachments/files/30561342/rerere-timed-client.tar.gz)

## Summary by Sourcery

Introduce a dual Kubernetes client setup with read timeouts for non-watch operations while preserving non-timed clients for watch-based controllers and tests, and update logging, controllers, tests, and CI accordingly.

New Features:
- Add a timed Kubernetes client with a 30-second read timeout for request/response API calls and expose it via the shared library.
- Introduce a KubeClients struct to pass both timed and watch-safe clients throughout the operator, services, test utilities, and virtual machine helpers.
- Extend the reference value image controller to own PCR-compute jobs so that ApprovedImage status and related resources reconcile correctly on job changes.

Bug Fixes:
- Ensure machine cleanup failures in the keygen controller surface as finalizer cleanup errors instead of being silently ignored.
- Use dedicated watch clients for long-lived streams and resource wait helpers to avoid hangs and improve reliability of operator controllers and integration tests.

Enhancements:
- Refactor operator, attestation-key-register, register-server, compute-pcrs, and test utilities to consistently use the new timed client for non-watch operations and a separate client for watches.
- Add timestamps to test logging and enable env_logger humantime formatting for clearer, time-correlated logs.
- Replace ad-hoc eprintln error logging with structured warn logging in the attestation key register secret reconciliation logic.
- Increase certain namespace and resource cleanup timeouts in tests to reduce flakiness on slower clusters.
- Adjust KubeVirt VM helpers to use the appropriate timed or watch clients for resource creation and readiness checks.

CI:
- Add a must-gather collection and artifact upload step to the GitHub Actions integration test workflow to aid in debugging CI failures.

Tests:
- Update integration and attestation tests to use separate timed and watch clients, including for await_condition and wait_for_resource_deleted calls, to improve stability across platforms.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -76,3 +76,14 @@ jobs:
         run: |
           eval $(ssh-agent -s)
           make integration-tests
+      - name: "Gather must-gather"
+        if: always()
+        run: must-gather/gather
+        env:
+          COLLECTION_PATH: must-gather-output
+      - name: "Upload must-gather"
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: must-gather-${GITHUB_HEAD_REF}
+          path: must-gather-output/
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -27,6 +27,31 @@ use anyhow::{Context, Result, anyhow};
 use conditions::*;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::{Condition, OwnerReference, Time};
 use kube::{Api, Client, Resource};
+use std::time::Duration;
+
+/// Read timeout for non-watch Kubernetes API calls.
+pub const KUBE_READ_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Create a [`Client`] with [`KUBE_READ_TIMEOUT`] applied.
+///
+/// Use this for request/response API calls. For long-lived watch streams, use
+/// [`Client::try_default`] instead so the connection never times out on idle.
+pub async fn timed_client() -> Result<Client> {
+    let mut config = kube::Config::infer().await?;
+    config.read_timeout = Some(KUBE_READ_TIMEOUT);
+    Ok(Client::try_from(config)?)
+}
+
+/// A pair of Kubernetes clients
+///
+/// `client` has a read timeout for request/response API calls.
+/// `watch` has no read timeout, keeping it safe for long-lived watch streams
+/// (Controllers, Reflectors, `await_condition`).
+#[derive(Clone)]
+pub struct KubeClients {
+    pub request_client: Client,
+    pub watch_client: Client,
+}
 
 #[macro_export]
 macro_rules! update_status {
--- a/operator/src/attestation_key_register.rs
+++ b/operator/src/attestation_key_register.rs
@@ -25,7 +25,7 @@ use kube::{
         watcher,
     },
 };
-use log::info;
+use log::{info, warn};
 use serde_json::json;
 use std::{collections::BTreeMap, sync::Arc, time::Duration};
 
@@ -335,7 +335,7 @@ async fn secret_reconcile(
                     .await
                     .map(|_| Action::await_change())
                     .map_err(|e| {
-                        eprintln!("Error updating attestation key volumes on secret apply: {e}");
+                        warn!("Error updating attestation key volumes on secret apply: {e}");
                         finalizer::Error::<ControllerError>::ApplyFailed(e.into())
                     })
             }
@@ -349,9 +349,7 @@ async fn secret_reconcile(
                     .await
                     .map(|_| Action::await_change())
                     .map_err(|e| {
-                        eprintln!(
-                            "Error updating attestation key volumes during secret deletion: {e}"
-                        );
+                        warn!("Error updating attestation key volumes during secret deletion: {e}");
                         finalizer::Error::<ControllerError>::CleanupFailed(e.into())
                     })
             }
--- a/operator/src/attestation_key_register.rs
+++ b/operator/src/attestation_key_register.rs
@@ -213,7 +213,7 @@ async fn machine_reconcile(
     // Check if the machine is being deleted
     if machine.metadata.deletion_timestamp.is_some() {
         info!(
-            "Machine {} is being deleted, updating attestation key volumes",
+            "Machine {} is being deleted, skipping update of attestation key volumes",
             machine.metadata.name.clone().unwrap_or_default()
         );
         return Ok(Action::await_change());
--- a/operator/src/attestation_key_register.rs
+++ b/operator/src/attestation_key_register.rs
@@ -30,8 +30,8 @@ use serde_json::json;
 use std::{collections::BTreeMap, sync::Arc, time::Duration};
 
 use trusted_cluster_operator_lib::conditions::ATTESTATION_KEY_MACHINE_APPROVE;
-use trusted_cluster_operator_lib::endpoints::*;
-use trusted_cluster_operator_lib::{AttestationKey, AttestationKeyStatus, Machine, update_status};
+use trusted_cluster_operator_lib::{AttestationKey, AttestationKeyStatus, KubeClients, Machine};
+use trusted_cluster_operator_lib::{endpoints::*, update_status};
 
 use crate::conditions::attestation_key_approved_condition;
 use crate::trustee;
@@ -44,6 +44,7 @@ use operator::{
 /// Stores give local cache access to avoid repeated API-server reads.
 pub struct AkContextData {
     pub client: Client,
+    watch_client: Client,
     pub machine_store: Store<Machine>,
     pub ak_store: Store<AttestationKey>,
     pub secret_store: Store<Secret>,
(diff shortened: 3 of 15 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -76,3 +76,14 @@ jobs:
...
+      - name: "Gather must-gather"
+        if: always()
+        run: must-gather/gather
+        env:
+          COLLECTION_PATH: must-gather-output
+      - name: "Upload must-gather"
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: must-gather
```
SpaceFace02: Maybe we can differentiate each GHA must-gather run by some unique identifier? Not sure if GHA adds a unique id based on which workflow created the file.

.github/workflows/integration-tests.yml
```diff
@@ -76,3 +76,14 @@ jobs:
...
+          path: must-gather-output/
```
SpaceFace02, line 89: Any way to configure retention period for this, otherwise these files may pile up to use a lot of space.

operator/src/register_server.rs
```diff
@@ -179,6 +179,10 @@ async fn keygen_reconcile(
                             );
                             return Ok(Action::await_change());
                         }
+                        Err(e) => {
+                            let err = anyhow!("{e}").into();
```
SpaceFace02: Q: When you say GET request to TEC fails occasionally, do you mean a request timeout, or a 4xx/5xx error which is returned? Why does it fail (due to flakiness or other reason)?

lib/src/lib.rs
```diff
@@ -27,6 +27,31 @@ use anyhow::{Context, Result, anyhow};
...
+}
+
+/// A pair of Kubernetes clients
+///
+/// `client` has a read timeout for request/response API calls.
+/// `watch` has no read timeout, keeping it safe for long-lived watch streams
+/// (Controllers, Reflectors, `await_condition`).
+#[derive(Clone)]
+pub struct KubeClients {
+    pub client: Client,
```
SpaceFace02: nit: Like Sourcery mentioned, the naming could be changed to something less generic.

operator/src/attestation_key_register.rs
```diff
@@ -213,7 +216,7 @@ async fn machine_reconcile(
     // Check if the machine is being deleted
     if machine.metadata.deletion_timestamp.is_some() {
         info!(
-            "Machine {} is being deleted, updating attestation key volumes",
+            "Machine {} is being deleted, skipping update of attestation key volumes",
```
SpaceFace02, line 219: We are skipping update of attestation key volumes because as per the latest PR by Roy, attestation keys are no longer stored in volumes, but are stored in reference values?

If so, skipping updates might be a misnomer (because there is no attestationkey volume which can be skipped anyways). Ideally we should have some job that deletes that short lived attestationKey from trustee reference values)

lib/src/lib.rs
SpaceFace02: nit: Also I don't know what would be the ideal way to name this, whether it should have the 's' in the end, indicating a plural. It is one object after all, with different functionalities available

SpaceFace02 (commented):
Thanks for the PR! I have a few small suggestions

.github/workflows/integration-tests.yml
Jakob-Naucke: Good idea, I've added the source branch. We could add the artifact ID but IMO the naming you get from the browser downloads (e.g. `must-gather-timed-client.zip`, `must-gather-timed-client.zip(1)`) is more intuitive.

.github/workflows/integration-tests.yml
Jakob-Naucke: For public GitHub projects like this one, you get 90 days. It's possible to set something lower here, but as long as GitHub lets us I don't really see a reason to. FWIW one of these is about 120K zipped when all went well, about 750K with failures, so at 30 a week, I don't really see the total size exceeding 500M which is less than what you get with e.g. a free Dropbox account. (The size of OpenShift CI artifacts is another story… 🥲)

operator/src/register_server.rs
Jakob-Naucke: The request would be launched and never return because prior to this PR, it has no timeout. From a log with UUID per request, which is printed again upon return:

```
2026-07-28T15:32:33Z GET /api/v1/namespaces/test-4a79467d/secrets/trustee-auth "" 46a1e037-ff78-4a05-97b4-ecefc2ac314d
2026-07-28T15:32:33Z 46a1e037-ff78-4a05-97b4-ecefc2ac314d
```

and then this GET never returning occurred often:

```
[2026-07-28T15:33:19Z INFO  operator::register_server] get TEC
2026-07-28T15:33:19Z GET /apis/trusted-execution-clusters.io/v1alpha1/namespaces/test-4a79467d/trustedexecutionclusters/trusted-execution-cluster "" 1426a6a1-91fb-4d4f-856d-5871491635d8
```

(with that UUID never seen again, despite the log continuing until 15:38)

operator/src/attestation_key_register.rs
Jakob-Naucke: > If so, skipping updates might be a misnomer

yes, and this comment is not touched in #248, could you raise that there?

Jakob-Naucke (commented):
> Several paths now clone full KubeClients instances when only one client is used (e.g. in launch_keygen_controller, launch_rv_*_controller, and test helpers); you could reduce unnecessary cloning by passing references or extracting just the needed client to keep ownership and lifetimes simpler.

@sourcery-ai these all use both

.github/workflows/integration-tests.yml
SpaceFace02: Yes, adding the source branch would make it a lot clearer and less prone to accidental deletion.

.github/workflows/integration-tests.yml
SpaceFace02: Ohh, If Github automatically deletes the artifacts after 90 days, then no issues regarding storage running out, although we might need to save the important logs manually (if for whatever reason we need to look at log history >90 days), but that's a completely different issue of its own.

How much storage does Openshift CI take for 1 integration test run on Azure?

operator/src/register_server.rs
SpaceFace02: So the TEC response never returns due to the TEC still being in the process of initializing when the GET request was made? And the subsequent requeue (after merging this PR) replays the request once TEC is ready?

.github/workflows/integration-tests.yml
Jakob-Naucke: between 300M and 1G gzipped (1G and 2G uncompressed) :(

operator/src/register_server.rs
Jakob-Naucke: > So the TEC response never returns due to the TEC still being in the process of initializing when the GET request was made?

There is no indication of TEC initializing; last reconciliation on TEC in that log is prior and `await_change`:
```
[2026-07-28T15:28:27Z INFO  operator] reconciled (ObjectRef { dyntype: (), name: "trusted-execution-cluster", namespace: Some("test-4a79467d"), extra: Extra { resource_version: Some("3636"), uid: Some("606aa85b-fadd-49c0-a8b8-a928fcafd962") } }, Action { requeue_after: None })
```

I believe this is general sporadic network failure that an operator should just defend against.

> And the subsequent requeue (after merging this PR) replays the request once TEC is ready?

It replays it, after 60s:
https://github.com/trusted-execution-clusters/operator/blob/429f0702d2bf91bf23440dd24002fee1f291024e/operator/src/lib.rs#L34-L37

SpaceFace02: Great, sounds good

SpaceFace02 (dismissed):
LGTM 👍 

> nit: Also I don't know what would be the ideal way to name this, whether it should have the 's' in the end, indicating a plural. It is one object after all, with different functionalities available

Minor nit (but that shouldn't stop this PR from getting merged)

lib/src/lib.rs
```diff
@@ -27,6 +27,31 @@ use anyhow::{Context, Result, anyhow};
...
+    let mut config = kube::Config::infer().await?;
+    config.read_timeout = Some(KUBE_READ_TIMEOUT);
+    Ok(Client::try_from(config)?)
+}
+
+/// A pair of Kubernetes clients
+///
+/// `client` has a read timeout for request/response API calls.
+/// `watch` has no read timeout, keeping it safe for long-lived watch streams
+/// (Controllers, Reflectors, `await_condition`).
```
alicefr, line 49: can you please document when one should use one or the other, it isn't very clear from the comment

uril:
Looks good to me, thanks.

alicefr:
A summary of the discussion we had offline with Jakob. Reading the [kube-rs documentation](https://docs.rs/kube/latest/kube/struct.Config.html) more carefully, it seems that the read_timeout only influences the direct API calls not the watches. (see also [this PR](https://github.com/kube-rs/kube/pull/1945)).  Hence, we don't need the 2 clients. 
Additionally, in the past, we saw also patch (write) operations hanging, so it makes sense also to set the write_timeout as well

PR #333, opened 2026-08-11
dependabot: Block cert-manager

Jakob-Naucke:
With our goal to stay on k8s 1.35 for easier downstreaming, cert-manager must also be blocked or 1.36 will be pulled as transitive dependency.

## Summary by Sourcery

Build:
- Update Dependabot configuration to restrict cert-manager version upgrades to 1.21 or higher without advancing Kubernetes beyond 1.35.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -36,6 +36,8 @@ updates:
         versions: [">=0.24"]
       - dependency-name: "sigs.k8s.io/controller-tools"
         versions: [">=0.21"]
+      - dependency-name: "github.com/cert-manager/cert-manager"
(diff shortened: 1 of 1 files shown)


SpaceFace02:
LGTM

Had 1 q: What is Openshift CI's update cadence? We could track a task to update Openshift CI if an update is available

SpaceFace02 (approved):

Jakob-Naucke:
> Had 1 q: What is Openshift CI's update cadence? We could track a task to update Openshift CI if an update is available

I'm not sure where one would hear that first :/

PR #334, opened 2026-08-11
Prepare tests for downstream

Jakob-Naucke:
- Allow for Makefile to be included from elsewhere
- Allow for setting an upstream directory other than `..`
- Make `get_encoded_root_pem`, `get_env` public
- Make `TestContext` clonable
- Use `TestContext.info` when possible to save lines
- a rename

Preview of downstream test to use these changes: https://github.com/Jakob-Naucke/confidential-clusters-operator/tree/scale-test

## Summary by Sourcery

Prepare test utilities and build tooling to be reusable from downstream projects and support multiple virtualization backends.

New Features:
- Allow configuring the upstream workspace directory via the UPSTREAM_DIR environment variable when generating operator manifests.
- Expose helper functions for environment variable access and root CA PEM encoding for reuse across modules and downstream tests.
- Introduce a shared NodeBackend trait to abstract SSH execution and root key retrieval across different VM backends.

Enhancements:
- Make TestContext clonable and consolidate logging through its info method to reduce verbosity in tests and utilities.
- Refactor VM backend configuration to pass an already encoded CA PEM into ignition generation instead of encoding it per backend.
- Update kubevirt and Azure backends to implement the new NodeBackend trait and use the shared shell execution helper.
- Adjust internal APIs (e.g., get_root_key signature and ssh_exec renamed to sh_exec) to better fit shared backend abstractions.

Build:
- Make the Makefile self-contained by resolving paths relative to the Makefile directory, enabling inclusion from other repositories and fixing version resolution for Go and Cargo tools.
- Update trusted-cluster-gen build target to work correctly when the Makefile is invoked from external directories.

Tests:
- Adapt attestation tests to the updated VM backend interface, passing Kubernetes client and namespace to retrieve root encryption keys.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,13 +7,18 @@
 
 SHELL := /bin/bash
 
+# Define directory of this Makefile so it can be `include`d from
+# elsewhere without variables breaking, e.g. for use of controller-gen
+# & kopium from this directory's LOCALBIN.
+MAKEFILE_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+
 NAMESPACE ?= trusted-execution-clusters
 PLATFORM ?= kind
 
 KUBECTL=kubectl
 INTEGRATION_TEST_THREADS ?= 1
 
-LOCALBIN ?= $(shell pwd)/bin
+LOCALBIN ?= $(MAKEFILE_DIR)/bin
 # either linux or darwin
 OS ?= $(shell uname -s | tr '[:upper:]' '[:lower:]')
 # either x86_64/amd64 or aarch64/arm64
@@ -28,11 +33,11 @@ else ifeq ($(OS),darwin)
   KOPIUM_TARGET := $(KOPIUM_RUST_ARCH)-apple-darwin
 endif
 
-CONTROLLER_TOOLS_VERSION ?= $(shell go list -m -f '{{.Version}}' sigs.k8s.io/controller-tools)
+CONTROLLER_TOOLS_VERSION ?= $(shell cd $(MAKEFILE_DIR) && go list -m -f '{{.Version}}' sigs.k8s.io/controller-tools)
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen-$(CONTROLLER_TOOLS_VERSION)
-YQ_VERSION ?= $(shell go list -m -f '{{.Version}}' github.com/mikefarah/yq/v4)
+YQ_VERSION ?= $(shell cd $(MAKEFILE_DIR) && go list -m -f '{{.Version}}' github.com/mikefarah/yq/v4)
 YQ ?= $(LOCALBIN)/yq-$(YQ_VERSION)
-KOPIUM_VERSION ?= $(shell cargo metadata --format-version 1 | jq -r '.resolve.nodes[] | select(.deps[]?.name == "kopium") | .deps[] | select(.name == "kopium") | .pkg | split("@")[1]')
+KOPIUM_VERSION ?= $(shell cd $(MAKEFILE_DIR) && cargo metadata --format-version 1 | jq -r '.resolve.nodes[] | select(.deps[]?.name == "kopium") | .deps[] | select(.name == "kopium") | .pkg | split("@")[1]')
 KOPIUM ?= $(LOCALBIN)/kopium-$(KOPIUM_VERSION)
 
 REGISTRY ?= quay.io/trusted-execution-clusters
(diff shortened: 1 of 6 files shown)


test_utils/src/lib.rs
```diff
@@ -457,8 +455,19 @@ pub async fn get_cluster_url(
         .await
 }
 
+pub async fn get_encoded_root_pem(client: Client, namespace: &str) -> Result<String> {
+    let secrets: Api<Secret> = Api::namespaced(client.clone(), namespace);
+    let root_secret = secrets.get(ROOT_SECRET).await?;
+    let root_secret_data = root_secret.data.unwrap();
```
SpaceFace02: Might want to use unwrap_or_else, or in general, improve error handling

Makefile
```diff
@@ -3,20 +3,19 @@
...
 
-.PHONY: all build build-tools crds-rs generate manifests cluster-up cluster-down \
-	install-trustee install clean fmt-check clippy lint test test-release release-tarball prepare-release \
-	operator-image compute-pcrs-image reg-server-image attestation-key-register-image image \
-	push-operator push-compute-pcrs push-reg-server push-attestation-key-register push \
+.PHONY: all build build-tools crds-rs generate manifests cluster-up cluster-down image push install-trustee install clean fmt-check clippy lint test test-release release-tarball prepare-release
 
 SHELL := /bin/bash
 
+MAKEFILE_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
```
SpaceFace02, line 13: Can you validate and confirm no other targets need this path change, like `crds-rs` target or `build` etc

SpaceFace02:
nit: One of your commits name is `[test_utils/virt: s/s(sh_exec)/\1/]`. is this intentional?

Makefile
Jakob-Naucke: `make build`, `make crds-rs` still work but I'm not sure what your question is?

Jakob-Naucke (commented):
> nit: One of your commits name is `[test_utils/virt: s/s(sh_exec)/\1/]`. is this intentional?

Yes, it's the sed syntax for replacing `ssh_exec` with `sh_exec`

Makefile
SpaceFace02: I meant if this change allows us to execute the make target commands from anywhere, `build`, `crds-rs` and other targets need to also be executed in the same context as the Makefile,
Basically don't we need the `cd $(MAKEFILE_DIR)` commands for other targets like `build` or `crds-rs`?

SpaceFace02:
> Yes, it's the sed syntax for replacing ssh_exec with sh_exec

Right :)

~~Can you confirm that the special characters in the commit message wouldn't cause issues in the CI or downstream? (It probably won't)~~

NVM, looks like the only disallowed character in a commit message is a NULL byte.

Makefile
Jakob-Naucke: Good catch; no, none of those work. I've updated the downstream Makefile to include a target for trusted-cluster-gen, and changed the commit here to only cover variables, not targets because otherwise it's a bit misleading. Or we would have all the `cd $(MAKEFILE_DIR)` but it's unnecessarily invasive IMO.

SpaceFace02: Ah, I see,

Could you please share the link to the downstream Makefile, I'm not sure if its [this](https://github.com/trusted-execution-clusters/operator/blob/6a744d7ec467224319d1443f132ce1bba6c78ac3/Makefile)?

Jakob-Naucke: Ah yes. It's unmerged but what I'd propose is https://github.com/Jakob-Naucke/confidential-clusters-operator/blob/4955ba4a30450898f51c2a70ece2a462371b4aef/tests/Makefile

SpaceFace02: Ahh, looks good, 

Can you please add a comment outlining this, so its clearer?

Jakob-Naucke:
ugh, lief. I also didn't know you'd have the test with your PR once you'd run it

/test azure-integration-test

Jakob-Naucke:
still lief, but it seems better now

/test azure-integration-test

Jakob-Naucke:
still

/test azure-integration-test

Jakob-Naucke:
@SpaceFace02 I'd live without passing Azure tests, does the comment LGTY?

Jakob-Naucke:
Seemed like a delayed Trustee pod restart, so this should improve with #248 

/test azure-integration-test

